Issue #13, opened 2018-02-24
Merge with Reconstructables.jl


jw3126:
@tkf I think it would be nice to have only a single package for modifying immutable values. I propose to merge `Reconstructables` and `Setfield`.
Do you agree with this goal in principle? If so we can work out here, which features we want exactly.

tkf:
I think that's an awesome idea!

For that, I'd suggest to add a `@settable` macro for custom inner constructor like this:

```julia
@settable struct S{X, Y}
    x::X
    y::Y

    S(x::X; y::Y=nothing) where {X, Y} = <function body>
end
```

expanding into

```julia
struct S{X, Y}
    x::X
    y::Y

    S(x::X; y::Y=nothing) where {X, Y} = <function body>
    S(x::X, y::Y=nothing) where {X, Y} = <function body>  # positional-only version
end
```

Or `@add_posonly` or `@add_kwonly`.

But first let's see how the discussion on https://discourse.julialang.org/t/ann-setfield-jl-yet-another-package-for-working-with-immutables/9267/ goes

tkf:
I also wonder if it can go into https://github.com/JuliaCollections

jw3126:
If we rely on a positional constructor and had the `@setable` macro, would that cover all your usecases of `Reconstructables.jl` conveniently? In order to apply `@setable`, you need to own the type. So if there was a type that
1) has no positional constructor
2) has a keywords only constructor
3) is not owned by the user

`Reconstructables.jl` would be more convenient. Does this situation arise in practice?

tkf:
Yes, I think `@settable` will cover all use-cases I have in mind (and more).  I think `DEProblem` subtypes in DiffEqBase.jl are good immutable struct examples.  The package is not too complex but it is a basis of the large DifferentialEquations.jl ecosystem so it's a good specimen:
https://github.com/JuliaDiffEq/DiffEqBase.jl/tree/master/src/problems
I'm sure there are more but that's I can think of at the moment.

The three conditions you mentioned for keyword-only constructor is the only working option seems to be unreasonable.  I don't think there is a way to solve set-ability problem for struct types you don't own without language support and without type piracy.  And I agree the probability that a struct has positional-only constructor is higher.

[...]

PR #17, opened 2018-03-01
Settable

jw3126:
#13 
@tkf can you play around with this branch and see if it fits your needs?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.6
+MacroTools
--- a/src/Setfield.jl
+++ b/src/Setfield.jl
@@ -24,4 +24,6 @@ end
 
 include("lens.jl")
 include("sugar.jl")
+include("macrotools.jl")
+include("settable.jl")
 end
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -117,7 +117,9 @@ end
 #     ret
 # end
 
-constructor_of(::Type{T}) where {T} = T
+@generated constructor_of(::Type{T}) where T =
+    getfield(T.name.module, Symbol(T.name.name))
+constructor_of(T::UnionAll) = constructor_of(T.body)
 
 @generated function setproperty(obj, ::Val{name}, val) where {name}
     T = obj
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -1,4 +1,7 @@
 using MacroTools
+
+const STRUCTSYMBOL = VERSION < v"0.7-" ? :type : :struct
+
 function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
@@ -45,6 +48,7 @@ function splittypedef(ex)
     d
 end
 
+
 function combinetypedef(d)
     name = d[:name]
     parameters = d[:params]
@@ -58,7 +62,8 @@ function combinetypedef(d)
         $(fields...)
         $(d[:constructors]...)
     end
-    Expr(:type, d[:mutable], header, body)
+
+    Expr(STRUCTSYMBOL, d[:mutable], header, body)
 end
 
 function combinefield(x)
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -6,18 +6,9 @@ function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
 
-function statements(ex)
-    exprs = []
-    if isexpr(ex, :block)
-        append!(exprs, vcat(map(statements, ex.args)...))
-    else
-        push!(exprs, ex)
-    end
-    return exprs
-end
-
 function splittypedef(ex)
     ex = MacroTools.striplines(ex)
+    ex = MacroTools.flatten(ex)
     d = Dict{Symbol, Any}()
     if @capture(ex, struct header_ body__ end)
         d[:mutable] = false
@@ -52,7 +43,7 @@ function splittypedef(ex)
         elseif item isa Symbol
             push!(d[:fields], (item, Any))
         else
-            append!(d[:constructors], statements(item))
+            push!(d[:constructors], item)
         end
     end
     d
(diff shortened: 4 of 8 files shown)


tkf:
Cool! I'm on trip right now. So it'll take a few more days to try it out.

tkf:
I suggest to generate the body of positional-only constructor from user-defined inner constructor (implementation in #18).

This helps for enforcing some invariants, e.g.:

> ```julia
> julia> struct OrderedPair
>            x::Real
>            y::Real
>            OrderedPair(x,y) = x > y ? error("out of order") : new(x,y)
>        end
> ```
> --- https://docs.julialang.org/en/latest/manual/constructors/#Inner-Constructor-Methods-1

jw3126:
This version also adds the body of the first constructor to the positional constructor. So invariants in the body of the first constructor are enforced.

tkf:
Oooh, yes, sorry, you are right.  I thought I had a good reason when started writing a patch...

So the main difference between #17 and #18 is that I defined the positional-only inner constructor solely based on user-defined inner constructor in #18.  On the other hand, your `default_constructor_dict` in #17 generates function signature (args, kwargs, params and whereparams) from field and struct type parameters.  I think this way of generating positional-only constructor is limiting, since:

(1) User may write some "type restriction" in `where` part of the constructor.  We will miss it if we use the ones from struct `where`.  On the other hand, the `where` in struct would be imposed when invoking `new` so we don't need to worry about it.

(2) User may use different type parameter names in constructor than in struct type parameters.  It is awful coding habit, but we need to have a way to detect and abort for such case than runtime cryptic `UndefVarError` errors, if we go with #17's `default_constructor_dict`.  #18 would allow such code so it is in a way more robust and we don't need to write extra code to rule out such bad code.

(3) User may add a custom promotion rule inside the struct.  It will be incompatible with the function signature derived from struct type parameters.  Example:

```julia
struct RangeWithInitial{X}
    xmin_xmax::Tuple{X,X}
    x0::X
    function RangeWithInitial(xmin_xmax::Tuple, x0)
        xmin, xmax, x0 = promote(xmin_xmax..., x0)
        return new{typeof(x0)}((xmin, xmax), x0)
    end
end
```

#### Related suggestion to default `constructor_of` definition

I think it is highly likely that people write constructor without "`{}`"

```julia
@settable struct ABC{A,B,C}
    a::A
    b::B
    c::C
    function ABC(a::A; b::B = 1, c::C = 2) where {A, B, C}
        ... some code ...
        return new{A, B, C}(a, b, c)
    end
end
```

than with "`{}`":

```julia
@settable struct ABC{A,B,C}
    ...
    function ABC{A,B,C}(a::A; b::B = 1, c::C = 2) where {A, B, C}
        #       ^^^^^^^
        ...
    end
end
```

In current status of Setfield.jl, without-`{}` constructor needs custom `constructor_of`:

```julia
constructor_of(::Type{<: ABC}) = ABC
```

to make `setproperty` work.  So I suggest to modify default `constructor_of` to:

```julia
constructor_of(T::Type) = getfield(T.name.module, Symbol(T.name.name))
constructor_of(T::UnionAll) = constructor_of(T.body)
```

to get the type-parameter -less version of the type constructor.  The default `setproperty` function then works with type-parameter -less constructor out-of-the-box.  Note that it works with default constructor since Julia generates type-parameter -less constructor as well.

The reason why I like this approach is that it's easy/trivial to get [“differentiable lens”](https://discourse.julialang.org/t/ann-setfield-jl-yet-another-package-for-working-with-immutables/9267/44?u=tkf) if `constructor_of` returns the type-parameter -less constructor.

tkf:
... as well as other situations in which changing type parameters makes sense:
https://discourse.julialang.org/t/ann-setfield-jl-yet-another-package-for-working-with-immutables/9267/11?u=tkf

tkf:
So, type-parameter -less constructor works, without breaking `@inferred` tests: #19

jw3126:
@tkf should we merge this into master or does it require further changes? Does the `@settable` macro here suffice to migrate your uses of `Reconstructables.jl` to `Setfield.jl`?

tkf:
Yes, I think so, in principle.  I haven't started using Reconstructables.jl yet, so I don't have a practical codebase to try it out.  (As actually I already started using Setfield.jl, since I realized lens was great!)  But I think our test covers important cases, so I think this PR is good to go.


### A minor point: Error/warn when `@settable` is a no-op?

We can worry about this later, but maybe it's good to throw in `@settable` when it is no-op?  I mean the test cases like:

```julia
@settable struct NoConstructor{A,B}
    b::B
end

@settable struct ExplicitConstructor{A,B}
    b::B
    ExplicitConstructor{A,B}(b::B) where {A,B} = new{A,B}(b)
end
```

I think the users who would put `@settable` there may be expecting something else than no-op to happen; otherwise they won't put `@settable` there.  There are three possibilities:

1. No-op (i.e., current `@settable`).

   * This could actually be useful when you are debugging inner-consturctor which (temporary) invoke no-op path of `@settable`.  It is cumbersome to remove `@settable`.

2. Warn.

   * Not sure if it is noticeable by novice users.
   * Has same merit as no-op; i.e., you don't have to temporary remove `@settable` when debugging.

3. Throw.

   * We can be sure that everyone notices it.
   * Cumbersome when debugging.


*If* we want to be *extremely* careful here and try to be forward-compatible, I'd go with the option 3 (throw) since that's the only option that is forward-compatible in case we want to switch to option 1 or 2.

jw3126:
The only "danger" with the no-op I see, is that new users could think `Setfield` requires to annotate each and every singe type definition. As you say error is cumbersome for debugging. I would be fine with a warning or just an info on the other hand. If we want to change the behavior in future, we can still add a depwarn in the no-op case.

jw3126:
I don't really have an opinion, should we add a warning, you decide :)

tkf:
If we support QuickTypes and alike as in #21, we need to make `@settable` no-op for many kinds of statements.  So let it be no-op, for consistency.  (What a great excuse for being lazy :smile:)

jw3126:
Ok lets merge this!

PR #18, opened 2018-03-06
Generate body of constructor from given expr

tkf:
suggestion for #17

### DIFF
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -6,18 +6,9 @@ function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
 
-function statements(ex)
-    exprs = []
-    if isexpr(ex, :block)
-        append!(exprs, vcat(map(statements, ex.args)...))
-    else
-        push!(exprs, ex)
-    end
-    return exprs
-end
-
 function splittypedef(ex)
     ex = MacroTools.striplines(ex)
+    ex = MacroTools.flatten(ex)
     d = Dict{Symbol, Any}()
     if @capture(ex, struct header_ body__ end)
         d[:mutable] = false
@@ -52,7 +43,7 @@ function splittypedef(ex)
         elseif item isa Symbol
             push!(d[:fields], (item, Any))
         else
-            append!(d[:constructors], statements(item))
+            push!(d[:constructors], item)
         end
     end
     d
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -5,7 +5,17 @@ const STRUCTSYMBOL = VERSION < v"0.7-" ? :type : :struct
 function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
-    
+
+function statements(ex)
+    exprs = []
+    if isexpr(ex, :block)
+        append!(exprs, vcat(map(statements, ex.args)...))
+    else
+        push!(exprs, ex)
+    end
+    return exprs
+end
+
 function splittypedef(ex)
     ex = MacroTools.striplines(ex)
     d = Dict{Symbol, Any}()
@@ -42,7 +52,7 @@ function splittypedef(ex)
         elseif item isa Symbol
             push!(d[:fields], (item, Any))
         else
-            push!(d[:constructors], item)
+            append!(d[:constructors], statements(item))
         end
     end
     d
--- a/src/settable.jl
+++ b/src/settable.jl
@@ -16,6 +16,25 @@ function argsymbol(ex::Expr)::Symbol
     end
 end
 
+"""
+    strip_default_value(s::Symbol)
+    strip_default_value(ex::Expr)
+
+Strip-off the value part when `ex` is a key-value expression.
+It is an identity function for `Symbol`.
+"""
+strip_default_value(s::Symbol) = s
+
+function strip_default_value(ex::Expr)
+    arg = if isexpr(ex, :kw)
+        ex.args[1]
+    else
+        ex
+    end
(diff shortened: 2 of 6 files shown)


src/settable.jl
```diff
@@ -4,32 +4,72 @@ macro settable(ex)
...
+        def = splitdef(constructor)
+        args = map(argsymbol, def[:args])
+        kwargs = map(argsymbol, def[:kwargs])
+        if fields[1:length(args)] == args &&
+                Set(fields[length(args)+1:end]) <= Set(kwargs)
+            newargs = copy(def[:args])
+            newkwargs = []
+            for a in def[:kwargs]
+                if argsymbol(a) in fields
+                    push!(newargs, a)
```
jw3126: Maybe we should strip default values from `newargs`. Otherwise, we are implicitly adding mutliple constructors,
e.g. `T(a,b=2) = ...` adds two constructors. These implicit constructors pollute the method table (no big deal), but might also overwrite other existing constructors and lead to subtle bugs.

jw3126:
```julia
struct ManyConstructors
    a
    b
    ManyConstructors(a) = new(a,1)
    ManyConstructors(a;b=2) = new(a,b)
```

test/test_settable.jl
```diff
@@ -1,10 +1,64 @@
...
+    end
+end
+Setfield.constructor_of(::Type{<: UntypedConstructor}) = UntypedConstructor
+
+@testset "UntypedConstructor" begin
+    s1 = UntypedConstructor(1, 0)
+    s2 = @set s1.b = -2
+    @test s2 === UntypedConstructor(1, b=-2)
+    @test_throws AssertionError @set s1.b = 2
 end
```
jw3126: These are good tests! There should also be a test, that mimics the `@add_kwonly` situation.

src/settable.jl
tkf: Thanks! Good catch!

jw3126:
A case that worked out of the box before, but does not work out of the box anymore is:
```julia
struct S{A,B}
    a::A
end

s = S{Int, :B}(1)
@set s.a = 2
```
But I think having your differentiation example work out of the box is absolutely worth it. Can you remove now unnecessary `constructor_of` from the tests?

test/test_settable.jl
```diff
@@ -82,3 +82,21 @@ end
...
     @test s2 === ManyConstructors(0, b=-2)
 end
+
+# Mimic @add_kwonly from Reconstructables.jl and DiffEqBase.jl.
+@settable struct WithKwOnly
+    a
+    b
+    WithKwOnly(a; b=2) = new(a, b)
+    WithKwOnly(; a=error("`a` is mandatory"), b=2) = new(a, b)
+end
```
jw3126: Can you also include the macroexpansion? Just hardcode everything:
```julia
macro kwonly(code)
esc(quote
    WithKwOnly(a; b=2) = new(a, b)
    WithKwOnly(; a=error("`a` is mandatory"), b=2) = new(a, b)
end)
end
```

src/settable.jl
jw3126: Ah sorry that example was not good. 
```julia
struct ManyConstructors
    a
    b
    ManyConstructors(a) = new(a,1)
    ManyConstructors(a;b=2) = new(a,b)
end
```
Already overwrites the first constructor without `@settable`.
```julia
struct ManyConstructors
           a
           b
           ManyConstructors(a) = new(a,1)
           ManyConstructors(;a=1,b=2) = new(a,b)
       end
```
is a better example.

test/runtests.jl
```diff
@@ -198,6 +198,17 @@ end
     @test m.a === m_inner_init
 end
 
+struct A{X, Y}
+    x::X
+    y::Y
+end
+
+@testset "type change during @set (default constructor_of)" begin
```
tkf: Is this the test you mentioned in https://github.com/jw3126/Setfield.jl/pull/19#issuecomment-371087033 ?

jw3126: Yes, this looks good.

src/macrotools.jl
```diff
@@ -42,7 +52,7 @@ function splittypedef(ex)
         elseif item isa Symbol
             push!(d[:fields], (item, Any))
         else
-            push!(d[:constructors], item)
+            append!(d[:constructors], statements(item))
```
jw3126: There is functionality for removing blocks in MacroTools. I think you could just add [flatten](https://github.com/MikeInnes/MacroTools.jl/blob/e762757a65d6a4b45c90a942df2026a3ffb65207/src/utils.jl#L331) as another preprocessing step, like the striplines above.

jw3126:
I think this looks good! We should merge it into the `settable` branch.

test/test_macrotools.jl
```diff
@@ -34,11 +34,13 @@ using Setfield: splittypedef, combinetypedef
...
-        ex |> Base.remove_linenums!
+    @test ex |> splittypedef |> combinetypedef |>
+        Base.remove_linenums! |> MacroTools.flatten ==
+        ex |> Base.remove_linenums! |> MacroTools.flatten
 
     constructors = splittypedef(ex)[:constructors]
     @test length(constructors) == 1
-    @test first(constructors) == :((S(a::A) where A) = new{A}())
+    @test first(constructors) ==
+        :((S(a::A) where A) = new{A}()) |> MacroTools.flatten
```
tkf: I needed to tweak test_macrotools.jl to use `MacroTools.flatten` in `splittypedef`.

tkf:
If you are fine with the change for using MacroTools.flatten, please go ahead and merge.

PR #19, opened 2018-03-07
Return type-less constructor from constructor_of

tkf:
suggestion for #17; extending #18

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -117,7 +117,9 @@ end
 #     ret
 # end
 
-constructor_of(::Type{T}) where {T} = T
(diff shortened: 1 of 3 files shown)


jw3126:
A case that worked out of the box before, but does not work out of the box anymore is:
```julia
struct S{A,B}
    a::A
end

s = S{Int, :B}(1)
@set s.a = 2
```
But I think having your differentiation example work out of the box is absolutely worth it. Can you remove now unnecessary `constructor_of` from the tests?

tkf:
Closing, since #18 now includes this branch.

PR #21, opened 2018-03-11
Interoperability

tkf:
- [x] merge #17 to master
- [ ] then rebase onto master
- [ ] use Kwonly.jl instead of Reconstructables.jl for `@add_kwonly` test, once https://github.com/JuliaLang/METADATA.jl/pull/13769 is merged

Not sure we want to actually merge this.  It's just for start discussing interoperability with other packages, especially QuickTypes.jl.

### DIFF
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -1,4 +1,7 @@
 using MacroTools
+
+const STRUCTSYMBOL = VERSION < v"0.7-" ? :type : :struct
+
 function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
@@ -45,6 +48,7 @@ function splittypedef(ex)
     d
 end
 
+
 function combinetypedef(d)
     name = d[:name]
     parameters = d[:params]
@@ -58,7 +62,8 @@ function combinetypedef(d)
         $(fields...)
         $(d[:constructors]...)
     end
-    Expr(:type, d[:mutable], header, body)
+
+    Expr(STRUCTSYMBOL, d[:mutable], header, body)
 end
 
 function combinefield(x)
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -2,8 +2,13 @@ using MacroTools
 
 const STRUCTSYMBOL = VERSION < v"0.7-" ? :type : :struct
 
+struct ParseError <: Exception
+    msg::String
+end
+Base.showerror(io::IO, ex::ParseError) = print(io, "ParseError: $(ex.msg)")
+
 function parse_error(ex)
-    throw(ArgumentError("Cannot parse typedefinition from $ex."))
+    throw(ParseError("Cannot parse typedefinition from $ex."))
 end
 
 function splittypedef(ex)
@@ -47,6 +52,18 @@ function splittypedef(ex)
         end
     end
     d
+end
+
+
+function is_trivial_struct(ex)
+    try
+        return isempty(splittypedef(ex)[:constructors])
+    catch err
+        if err isa ParseError
+            return false
+        end
+        throw()
+    end
 end
 
 
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -6,18 +6,9 @@ function parse_error(ex)
     throw(ArgumentError("Cannot parse typedefinition from $ex."))
 end
 
-function statements(ex)
-    exprs = []
-    if isexpr(ex, :block)
-        append!(exprs, vcat(map(statements, ex.args)...))
-    else
-        push!(exprs, ex)
-    end
-    return exprs
-end
-
 function splittypedef(ex)
     ex = MacroTools.striplines(ex)
+    ex = MacroTools.flatten(ex)
     d = Dict{Symbol, Any}()
     if @capture(ex, struct header_ body__ end)
         d[:mutable] = false
@@ -52,7 +43,7 @@ function splittypedef(ex)
         elseif item isa Symbol
             push!(d[:fields], (item, Any))
         else
-            append!(d[:constructors], statements(item))
+            push!(d[:constructors], item)
         end
     end
     d
--- a/src/macrotools.jl
+++ b/src/macrotools.jl
@@ -1,0 +1,67 @@
+using MacroTools
+function parse_error(ex)
+    throw(ArgumentError("Cannot parse typedefinition from $ex."))
+end
+    
+function splittypedef(ex)
+    ex = MacroTools.striplines(ex)
+    d = Dict{Symbol, Any}()
+    if @capture(ex, struct header_ body__ end)
+        d[:mutable] = false
+    elseif @capture(ex, mutable struct header_ body__ end)
+        d[:mutable] = true
+    else
+        parse_error(ex)
+    end
+    
+    if @capture header nameparam_ <: super_
+        nothing
+    elseif @capture header nameparam_
+        super = :Any
+    else
+        parse_error(ex)
+    end
+    d[:supertype] = super
+    if @capture nameparam name_{param__}
+        nothing
+    elseif @capture nameparam name_
+        param = []
+    else
+        parse_error(ex)
+    end
+    d[:name] = name
+    d[:params] = param
+    d[:fields] = []
+    d[:constructors] = []
+    for item in body
+        if @capture item field_::T_
+            push!(d[:fields], (field, T))
+        elseif item isa Symbol
+            push!(d[:fields], (item, Any))
+        else
+            push!(d[:constructors], item)
+        end
+    end
+    d
+end
+
+function combinetypedef(d)
+    name = d[:name]
+    parameters = d[:params]
+    nameparam = isempty(parameters) ? name : :($name{$(parameters...)})
+    header = :($nameparam <: $(d[:supertype]))
+    fields = map(d[:fields]) do field
+        fieldname, typ = field
+        :($fieldname::$typ)
+    end
+    body = quote
+        $(fields...)
+        $(d[:constructors]...)
+    end
+    Expr(:type, d[:mutable], header, body)
+end
+
+function combinefield(x)
+    fieldname, T = x
+    :($fieldname::$T)
+end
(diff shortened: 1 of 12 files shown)


tkf:
Since QuickTypes generates a struct and a bunch of functions, we need to support blocks in `@settable` and ignore non-struct statements, if we are to support it.  Also, using QuickTypes (especially `@qstruct_fp`) with Setfield is not straight forward since QuickTypes generates only "type-parameter -full" inner constructor.  That is to say,

```julia
@qstruct_fp Plane(nwheels, weight::Number; brand=:zoomba)
```

generates an inner constructor

```julia
function (::Type{Plane{T1, T2, T3}}){T1 <: Any, T2 <: Number, T3 <: Any}(
        nwheels=nothing, weight=nothing; brand=:zoomba)
    return new{T1, T2, T3}(nwheels, weight, brand)
end
```

and an outer constructor

```julia
Plane{T1 <: Any, T2 <: Number, T3 <: Any}(
        nwheels::T1=nothing, weight::T2=nothing; brand::T3=:zoomba) =
    Plane{T1, T2, T3}(nwheels, weight; brand=brand)
```

*If* QuickTypes generated a type parameter-less inner constructor, i.e.,

```julia
function Plane(nwheels::T1 = nothing,
               weight::T2 = nothing;
               brand::T3 = :zoomba) where {T1 <: Any, T2 <: Number, T3 <: Any}
    return new{T1, T2, T3}(nwheels, weight, brand)
end
```

in addition to (or instead of) the above inner constructor, then `@settable` would have worked seamlessly.  Not sure there is anything else we can do in Setfield.jl side.

This is the actual test code:

https://github.com/tkf/Setfield.jl/blob/9bcb9f8e639709b4b494f131c2b2c788d48541ac/test/test_quicktypes.jl#L94-L122

ping @cstjean

src/macrotools.jl
```diff
@@ -0,0 +1,90 @@
...
+function is_trivial_struct(ex)
+    try
+        return isempty(splittypedef(ex)[:constructors])
+    catch err
+        if err isa ParseError
+            return false
+        end
+        throw()
+    end
+end
```
jw3126: My plan was to move splitt/combine typedef file into a PR to `MacroTools`, once we think its good enough. And the `macrotools.jl` file was meant just for things that need to be in this PR. I should have added a comment at the top of the file, sorry. Could you move `is_trivial_struct` elsewhere?

src/settable.jl
```diff
@@ -0,0 +1,107 @@
...
+
+function posonly_constructor(dtype::Dict)
+    combinedef(posonly_constructor_dict(dtype))
+end
+
+function settable(ex)
+    is_trivial_struct(ex) && return ex
+    M = current_module()
+    ex = macroexpand(M, ex)
+    return _settable(ex)
```
jw3126: You could alternatively use `MacroTools.postwalk` to make each typedef burried in `ex` settable.

test/runtests.jl
```diff
@@ -202,18 +219,23 @@ struct B{T, X, Y}
...
+if Pkg.installed("StaticArrays") != nothing
+    include("test_staticarrays.jl")
     include("spaceship.jl")
 end
+
+if Pkg.installed("QuickTypes") != nothing
+    @testset "QuickTypes" begin include("test_quicktypes.jl") end
+end
+
+if Pkg.installed("Reconstructables") != nothing
```
jw3126: The `if Pkg.installed("StaticArrays")` was because installing `StaticArrays` on `v0.7` broke CI. I think installing `QuickTypes` is no problem, so you could remove the condition. Same applies for `Reconstructables` / `Kwonly` I guess.

src/macrotools.jl
tkf: I see.  Should I revert macrotools.jl to the previous state?  I also introduced `ParseError` but I think PR for exposing an exception type as a public API has lower chance to get into MacroTools.jl.

src/settable.jl
tkf: Cool. I should be learning how to use MacroTools :)

test/runtests.jl
tkf: It seems CI fails around QuickTypes test.
https://travis-ci.org/jw3126/Setfield.jl/jobs/351898265

It looks like QuickTypes failed in v0.7 (though it's 5 months ago)
https://travis-ci.org/cstjean/QuickTypes.jl/builds/281258251

Anyway, I'll use the conditional only for excluding v0.7 incompatible packages.

jw3126: Maybe we should do 
```julia
if VERSION <v"0.7-"
    Pkg.add("SomePackage")
    using SomePackage
    # do tests
end
```

cstjean:
> If QuickTypes generated a type parameter-less inner constructor, i.e.,

Could you please show me which constructor call the `@set` macro makes, that should work but doesn't? I.e. what is `@macroexpand @set x0.brand = 31` on this branch?

I haven't updated QuickTypes.jl for 0.7 yet...

jw3126:
Okay say `x0::X{T1,T2}`. Then `@set x0.brand=31` will at some point call `X(31, x0.other_field)`. Note that the parameters `T1, T2` were stripped. (We choose to do this, so that `@set` can change the type, see [here](https://discourse.julialang.org/t/ann-setfield-jl-yet-another-package-for-working-with-immutables/9267/44?u=jw3126) for an example where this is useful.)

Now this PR is about interoperability with customized types, which may not have positional constructors. The idea is to add a macro 
```julia
@settable struct X{T1,T2}
    ...
end
```
that makes sure, that there is a `X(brand, other_field)` positional constructor. For this macro to work, we rely on the existence of an inner constructor of the form `X(...)`, but `QuickTypes` only has one of the form `X{T1,T2}(...)`.

cstjean:
Would it help if I made the outer constructor an inner constructor?

jw3126:
Yes, that would help!

cstjean:
I just moved the outer constructor inside the type definition, on the current `master`. It wasn't entirely trivial, so please report any issue you find, and let me know if it allows interoperability with this PR.

jw3126:
@cstjean cool, this is very much appreciated! Thanks a lot! @tkf are you still working on this? If not are you okay, if I take over this branch?

jw3126:
@cstjean Mhh I checked out the latest master of `QuickTypes`, but the outer constructor is still outside, it seems. We would need
```julia
@macroexpand @qstruct_fp Plane3(nwheels, weight::Number; brand=:zoomba)
```
expand to
```julia
   ...
    struct Plane3{T1 <: Any, T2 <: Number, T3 <: Any} <: Any
        nwheels::T1
        weight::T2
        brand::T3
        begin  # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 200:
            function (::Type{Plane3{T1, T2, T3}}){T1 <: Any, T2 <: Number, T3 <: Any}(nwheels, weight; brand=:zoomba) # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 202:
                nothing # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 203:
                return new{T1, T2, T3}(nwheels, weight, brand)
            end
            function Plane3(nwheels::T1 = nothing,
               weight::T2 = nothing;
               brand::T3 = :zoomba) where {T1 <: Any, T2 <: Number, T3 <: Any}
               return new{T1, T2, T3}(nwheels, weight, brand)
            end
        end
    end
    ...
```
The important points are 
* All the constructors parameters are captured by the `where` clause
* It is an inner constructor
* Its arguments have the same names + order as the fields of `Plane3`. 

Instead current behavior is the following:
```julia
:($(Expr(:toplevel, quote 
    $(Expr(:meta, :doc))
    struct Plane3{T1 <: Any, T2 <: Number, T3 <: Any} <: Any
        nwheels::T1
        weight::T2
        brand::T3
        begin  # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 200:
            function (::Type{Plane3{T1, T2, T3}}){T1 <: Any, T2 <: Number, T3 <: Any}(nwheels, weight; brand=:zoomba) # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 202:
                nothing # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 203:
                return new{T1, T2, T3}(nwheels, weight, brand)
            end
        end
    end
end, :(Plane3{T1 <: Any, T2 <: Number, T3 <: Any}(nwheels::T1, weight::T2; brand::T3=:zoomba) = Plane3{T1, T2, T3}(nwheels, weight; brand=brand)), :(function QuickTypes.construct(::Type{Plane3}, nwheels, weight, brand) # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 218:
        nothing # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 219:
        Plane3(nwheels, weight; brand=brand)
    end), :(function Base.show(io::IO, obj::Plane3) # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 80:
        print(io, typeof(obj)) # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 81:
        write(io, "(") # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 83:
        begin 
            show(io, obj.nwheels)
            write(io, ", ")
        end
        begin 
            show(io, obj.weight)
            nothing
        end # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 87:
        write(io, "; ") # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 90:
        begin 
            write(io, "brand")
            write(io, "=")
            show(io, obj.brand)
            nothing
        end # /home/jan/.julia/v0.6/QuickTypes/src/QuickTypes.jl, line 94:
        write(io, ")")
    end), nothing)))

```

cstjean:
Ahhh, very sorry about that, it seems that I forgot to merge my branch into `master`. Will do when I get home tonight.

jw3126:
No problem. Thanks again, I really like `QuickTypes`, looking forward to use it in conjunction with this package!

cstjean:
Merged and pushed.

jw3126:
@cstjean It works now! At least the `Plane3` example does. I will test more thoroughly after easter and tell you if I find any issues. Thanks again for making this possible!

jw3126:
I merged this PR, it works fine with the latest QuickTypes master.

cstjean:
Already did a few days ago.

PR #33, opened 2018-06-13
Fix test_staticarrays.jl

tkf:
It looks like StaticArrays.jl v0.7.1 was released before merging #31 :)
https://github.com/JuliaArrays/StaticArrays.jl/releases/tag/v0.7.1

Probably that's why the PR tests were fine but the latest build failed.


jw3126:
Cool!

Issue #32, opened 2018-06-12
Do we need MutationPolicy?


tkf:
Continuing the discussion from https://github.com/jw3126/Setfield.jl/pull/31#issuecomment-396480792

> I thought it was useful for generic programming that supports immutable and mutable types simultaneously. But I never needed it in practice.

Same here :)  Meaning that I thought it'd be useful but have never used yet.  I'm hoping to use it soon for writing generic code for mutable arrays and `StaticArray`.  So I thought I'd fix it for 0.7.  But I'm not entirely sure if I can write such code without handling the difference manually in the end.

Also, for Setfield.jl to be usable in my code (especially with Julia 0.7), dot-fusions like `.=` and `.+=` also have to be implemented which would add more complexity.

> OTOH I implement custom lenses from time to time and then the `MutationPolicy` machinery just adds annoying complexity.

Do you mean it's cumbersome to write policy-aware lens?  Thinking out loud, how about putting the policy in `Lens` type?  This way, you can define policy-unaware lens.  Additional benefit would be that you can compose `Lens{EncourageMutation}` and `Lens{ForbidMutation}` in arbitrary nested way.

jw3126:
Let me get one step back first. If you want to modify an object
1. It is easy if the object is mutable
2. Hard if the object is immutable

So `Setfield` should initially solve the first problem. Now both are easy. But then I thought there is a third case:

3. Hard if the object may or may not be mutable in generic code

So I thought this case should be covered as well by `Setfield` and introduced the `MutationPolicy` mechanism. However I never really needed it and I think it is half baked. 

Concerning dot-fusion support, what do you need? Do you want to update things with `.=` that may or may not be mutable like in 3.? Or do you just need 2. for things like
```julia
v = @SVector randn(3)
@set v[1:2] .= 42
```
For now my preference would be to get rid of support for 3. completely and only reintroduce it (via your suggestion or `MutationPolicy` or something else) if there is an actual use case that can guide the design.

tkf:
[...]

Issue #23, opened 2018-03-11
Lens and derivatives


jw3126:
@tkf's [amazing gist](https://discourse.julialang.org/t/ann-setfield-jl-yet-another-package-for-working-with-immutables/9267/44?u=jw3126) shows how we can use lenses to compute partial derivatives.

Often I am in the situation, where I want (partial) gradients, hessians etc. instead. Optimizing a struct is a typical example.

### Basic case
We have
```julia
struct Widget{T <: Number}
    parameter1::T
    parameter2::T
end

cost(w::Widget) = ...
```
and want to optimize `cost`. However many algorithms won't accept the cost function as is,
but need `cost(::AbstractVector)`. Also they often need gradients+hessians in the same form as well.

So to use such algorithms, a way is needed to convert our struct to a vector and back.
#### Variations:
* Instead of a flat struct, one often has nested structs.
* Some parameters of the struct might be fixed, while others can be optimized.

Currently I write a lot of ad hoc functions to pack unpack the nested fields I need into vector form and back.

I think there should be a better way. I would like to just provide for each of my structs an
isomorphism (e.g. a special type of lens) with some type of vectors and the lens machinery should 
be able to do the following for me:
* It should be easy to define the isomorphism for a big new struct, in terms of the isomorphisms of its fields.
* If should be easy to construct lenses that focus on a selection of nested fields.

tkf:
This is interesting direction!  How about providing AbstractVector (or even AbstractArray) interface/wrapper given an original object (e.g., struct) and a vector/array of lenses?  Or is equivalent to what you mean by isomorphism?  I'm thinking something like this:

```julia
using Setfield

struct VectorInterface{T, O, LS <: Tuple} <: AbstractVector{T}
    obj::O
    lenses::LS
end

function VectorInterface(obj::O, lenses) where O
    orig = promote((get(l, obj) for l in lenses)...)
    lenses = tuple(lenses...)
    return VectorInterface{typeof(orig[1]),
                           O,
                           typeof(lenses)}(obj, lenses)
end

Base.length(vi::VectorInterface) = length(vi.lenses)
Base.getindex(vi::VectorInterface, i) = get(vi.lenses[i], vi.obj)
Base.setindex(vi::VectorInterface, x, i) =
    set(Setfield.compose(vi.lenses[i], (@lens _.obj)), vi, x)
Setfield.hassetindex!(::VectorInterface) = false


struct Widget
    x1
    x2
    x3
    y1
    y2
[...]

PR #34, opened 2018-07-28
Remove mutation support

jw3126:


.travis.yml
```diff
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
+  - 0.7
```
tkf: Let's add 1.0 too! #35 :tada:

tkf:
Do you have some release plan?  How about doing it after merging this?

FYI using this branch with my package passed its tests https://travis-ci.org/tkf/Bifurcations.jl/builds/414973794 (although my Setfield usage is very basic).

src/lens.jl
```diff
@@ -146,22 +135,48 @@ function get(l::ComposedLens, obj)
...
-    if applicable(setindex!, obj, val, l.indices)
-        setindex!(obj, val, l.indices...)
-    else
-        set(l, obj, val, ForbidMutation())
+set(l::IndexLens, obj, val) = setindex(obj, val, l.indices...)
+
+
+const NNamedTupleLens{N,s} = NamedTuple{s, T} where {T <: NTuple{N, Lens}}
+struct MultiPropertyLens{L <: NNamedTupleLens} <: Lens
+    lenses::L
```
tkf: Why not use just a tuple of lenses, rather than a named tuple?  With a named tuple, you can only support properties, right?  (Though maybe that was the intention as you named it Multi**Property**Lens?)  I mean, I wonder if it makes sense to support "vectorizing" lens of mixed kinds, like this:

```julia
obj0 = (a=1, b=2, c=3)
l = @lens(  # this is a made-up syntax
  _.a,      # get/set by property
  _[3],     # get/set by index
)
obj1 = set(l, obj0, (100, 300))
@assert obj1 == (a=100, b=2, c=300)
```

test/test_quicktypes.jl
```diff
@@ -69,7 +69,7 @@ end
 @testset "Pack" begin
     x = Pack([Cat(:Tama, 1), Cat(:Pochi, 2)])
 
-    @set! x.animals[2].nlegs = 5
+    x = @set x.animals[2].nlegs = 5
```
tkf: I think this part is not tested since I commented `test_quicktypes.jl` out in https://github.com/jw3126/Setfield.jl/pull/31#discussion_r194624674. I suggest to pull #36 first or cherry-pick it.

tkf: Ah, so the mutability tests fail there with Julia 1.0.  I guess this PR should go first.

tkf: So #37 is failing as expected.

tkf: So it is failing because `IndexLens` does not support `Array` anymore.  I implemented a fix in #40.

Another (non-) solution is to use tuples in test_quicktypes.jl: #39

src/lens.jl
jw3126: Motivation was that I had lots of code, which looks like
```julia
obj = @set obj.a = a1
obj = @set obj.b = b1
obj = @set obj.c = c1
```
And the compiler was often unable to make this as fast as some handwritten
```julia
obj = Obj(a1,b1,c1, obj.other_fields...)
```
I like your approach, but I think it is harder for the compiler. I need to play around with it. Anyway this is a separate concern from removing mutation policy. It was a mistake to add it to this branch and I will remove it.

jw3126:
Tagging a new release after removing mutation support sounds fine to me!

src/lens.jl
tkf: Yeah, I agree that it's nice to release a simple version now and do some design iteration.  For the multi-lens, let's continue discussion and design iteration in https://github.com/jw3126/Setfield.jl/issues/23#issuecomment-413039326

jw3126:
@tkf does this look good to you, anything missing?

tkf:
Yup, #39 was all what I wanted to add.  It's also great that MacroTools stuff goes upstream (congrats!).

I think it's ready to go!

jw3126:
I tagged a new release! Thanks for #39!

cstjean:
I didn't follow this package very closely; why was `@set!` removed?

jw3126:
There is some discussion here https://github.com/jw3126/Setfield.jl/issues/32

cstjean:
I haven't tried using this package yet, but I'd like to use it to modify our default configuration. It would be cleaner to write `@set! config.thresh = 0.1`. In fact, what would be cooler would be

```julia
@set! begin
   config.thresh = 0.1
   config.N = 20
end
```

or

```julia
@set! config begin
    thresh = 0.1
    N = 20
end
```

or `@set!(config, thresh=0.1, N=20)`

I can write these macros myself on top of `Setfield`, but it would be nice to have official support.

I don't know if `MutationPolicy` is relevant. I just want `@set! a.b.c = 2` to become `a = @set a.b.c = 2`. I think I get the general goal you're trying to accomplish with `MutationPolicy`, but I'm not sure it's worth the added complexity.

jw3126:
Ah okay `MutationPolicy` is irrelevant for this. Having a version of `@set!` that does no mutation sounds fine to me! I would be happy about a PR. We can further discuss syntax then.

PR #35, opened 2018-08-11
Test against Julia 1.0

tkf:
I checked it locally that it works.  So I think it makes sense to have it in the CI.

It's a PR to #34.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - osx
 julia:
   - 0.7
+  - 1.0
   - nightly
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,6 @@ git:
 #before_script: # homebrew for mac
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
-# uncomment the following lines to override the default test script
-script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -4,7 +4,8 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
+  - '0.7'
+  - '1.0'
```
tkf: Note that I have to do `- '1.0'` instead of `- 1.0` since the latter would be interpreted as a float by the YAML parser. I changed 0.7 too for symmetry :)

.travis.yml
```diff
@@ -25,9 +26,6 @@ git:
 #before_script: # homebrew for mac
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
-# uncomment the following lines to override the default test script
-script:
-  - julia -e 'Pkg.clone(pwd()); Pkg.build("Setfield"); VERSION < v"0.7-" && Pkg.add("StaticArrays"); Pkg.test("Setfield"; coverage=true)'
```
tkf: So for Julia 1.0 we need to remove this since Pkg is not imported by default anymore.  Besides, we don't need to manually install it since we already have it in `test/REQUIRE`.

.travis.yml
```diff
@@ -4,7 +4,8 @@ os:
   - linux
   - osx
```
tkf: Also, why not stop testing against macOS to increase CI throughput? Setfield.jl is a pure Julia package so shouldn't it behave identically across all OSes (unless Julia has some very specific bugs)?

tkf:
closed by #39

PR #40, opened 2018-08-14
Fix IndexLens for mutable arrays

tkf:
This is alternative to #39.  If we want to handle mutable arrays (such as `Array`) by always copying them, this PR provides an implementation for it.  This is consistent with how `nomut` branch handles `mutable struct`.

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -3,7 +3,7 @@ export @lens
 export set, get, modify
 
 import Base: get
-using Base: setindex, getproperty
+using Base: getproperty
 
 """
     Lens
@@ -146,8 +146,25 @@ struct IndexLens{I <: Tuple} <: Lens
 end
 
 get(l::IndexLens, obj) = getindex(obj, l.indices...)
-set(l::IndexLens, obj, val) = setindex(obj, val, l.indices...)
+set(l::IndexLens, obj, val) = _setindex(obj, val, l.indices...)
 
+@generated function _setindex(obj, val, indices...)
+    if hasmethod(Base.setindex, Tuple{obj, val, indices...})
+        setter = Base.setindex
+    else
+        setter = setindex_on_copy
+    end
+    quote
+        $setter(obj, val, indices...)
+    end
+end
+
+function setindex_on_copy(obj, val, indices...)
+    clone = similar(obj, promote_type(eltype(obj), typeof(val)))
+    copyto!(clone, obj)
+    setindex!(clone, val, indices...)
+    return clone
+end
 
 const NNamedTupleLens{N,s} = NamedTuple{s, T} where {T <: NTuple{N, Lens}}
 struct MultiPropertyLens{L <: NNamedTupleLens} <: Lens
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - osx
 julia:
   - 0.7
+  - 1.0
   - nightly
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,6 @@ git:
 #before_script: # homebrew for mac
 #  - if [ $TRAVIS_OS_NAME = osx ]; then brew install gcc; fi
 
-# uncomment the following lines to override the default test script
-script:
-  - julia -e 'Pkg.clone(pwd()); Pkg.build("Setfield"); VERSION < v"0.7-" && Pkg.add("StaticArrays"); Pkg.test("Setfield"; coverage=true)'
 after_success:
   # push coverage results to Coveralls
   - julia -e 'cd(Pkg.dir("Setfield")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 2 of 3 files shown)


src/lens.jl
```diff
@@ -146,8 +146,25 @@ struct IndexLens{I <: Tuple} <: Lens
 end
 
 get(l::IndexLens, obj) = getindex(obj, l.indices...)
-set(l::IndexLens, obj, val) = setindex(obj, val, l.indices...)
+set(l::IndexLens, obj, val) = _setindex(obj, val, l.indices...)
 
+@generated function _setindex(obj, val, indices...)
+    if hasmethod(Base.setindex, Tuple{obj, val, indices...})
```
tkf: I know `hasmethod` is not pure and therefore is not legit inside `@generated` in principle.  But I've been seeing code deviating from the principle and somehow magically working.  My _guess_ was that this implementation is OK as long as `_setindex` is called after `Base.setindex(obj, ...)` is defined.  Since our StaticArrays tests are passing, this implementation is fine?

jw3126: Can you show a place, where people use code like this? Generally such things can cause bugs depending on import order of packages. But I don't understand enough to judge how safe this case is. I would guess that you could break this by
```julia
using Setfield
using StaticArrays

# define your own StaticArray with custom setindex
```
OTOH if you don't make this a generated function, with the new compiler it might still perform very well. Have you tested?

tkf: No I don't know the working code base using exactly this.  I should test this and also do benchmark for the dynamic version.

BTW, what is the new compiler optimization you are talking about?  The Union-splitting https://julialang.org/blog/2018/08/union-splitting sounds very close but the branching information still has to sit at the type level, right?  Also, this blog post is mentioning that method lookup is still slow.

But even if `@generated` approach does not work, we can always have a dynamic version as a fallback and then manually implement fast non-branching version for popular types (like `Array`, `Tuple` and `SArray`).  Probably it is not hard to get > 90% coverage by implementing it for a few types.

tkf:
So my benchmark shows the static version is ~20 times faster.

```julia
@generated function setindex_dynamic(obj, val, indices...)
    quote
        if hasmethod(Base.setindex, $(Tuple{obj, val, indices...}))
            setter = Base.setindex
        else
            setter = setindex_on_copy
        end
        setter(obj, val, indices...)
    end
end

@generated function setindex_static(obj, val, indices...)
    if hasmethod(Base.setindex, Tuple{obj, val, indices...})
        setter = Base.setindex
    else
        setter = setindex_on_copy
    end
    quote
        $setter(obj, val, indices...)
    end
end

function setindex_on_copy(obj, val, indices...)
    clone = similar(obj, promote_type(eltype(obj), typeof(val)))
    copyto!(clone, obj)
    setindex!(clone, val, indices...)
    return clone
end


using BenchmarkTools
a = [1, 2, 3]
println("\n", "*** setindex_dynamic")
display(@benchmark setindex_dynamic($a, 1, 1))
println("\n", "*** setindex_static")
display(@benchmark setindex_static($a, 1, 1))

using InteractiveUtils
versioninfo()
```

Output:

```
*** setindex_dynamic
BenchmarkTools.Trial:
  memory estimate:  144 bytes
  allocs estimate:  2
  --------------
  minimum time:     473.071 ns (0.00% GC)
  median time:      476.987 ns (0.00% GC)
  mean time:        529.616 ns (5.03% GC)
  maximum time:     177.103 μs (99.55% GC)
  --------------
  samples:          10000
  evals/sample:     196
*** setindex_static
BenchmarkTools.Trial:
  memory estimate:  112 bytes
  allocs estimate:  1
  --------------
  minimum time:     25.980 ns (0.00% GC)
  median time:      28.185 ns (0.00% GC)
  mean time:        38.229 ns (17.35% GC)
  maximum time:     38.632 μs (99.82% GC)
  --------------
  samples:          10000
  evals/sample:     992Julia Version 1.0.0
Commit 5d4eaca0c9 (2018-08-08 20:58 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
```

jw3126:
Thanks for the example, I agree it is somewhat exotic. But this kind of thing comes up in REPL sessions. Is `setindex` on `Vector` important to you? I would prefer removing it over having a slow / slightly buggy version. But if it is important to you, we can support it.

tkf:
> Is setindex on Vector important to you?

Nope :)  It is just me trying to make test_quicktypes.jl pass.

> I would prefer removing it over having a slow / slightly buggy version. 

Totally agree.  Let's pull #39 then!

src/lens.jl
jw3126: I had better constant propagation in mind. And I thought `hasmethod` might compile down to `true` or `false`, but it does not. It does dynamically look if there is a method.

tkf:
I asked the question about `@generated` here https://discourse.julialang.org/t/can-i-use-hasmethod-inside-generated-function/13645?u=tkf

PR #41, opened 2018-08-16
Build documents in test phase

tkf:
Hi, thanks for the new release! :tada: 

Looking at the release build, I just noticed that the documents are not updated because Documenter was silently failing.  This is not great that we don't notice when documents are broken.  So I suggest to build it in the test phase and mark it as failure if documents are broken.

### DIFF
--- a/docs/src/intro.md
+++ b/docs/src/intro.md
@@ -21,12 +21,12 @@ SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 ```
 Lets update the captains name:
 ```jldoctest spaceship
-julia> s.captain.name = "JULIA"
+julia> s.captain.name = :JULIA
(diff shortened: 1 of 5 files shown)


docs/src/intro.md
```diff
@@ -21,20 +21,20 @@ SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 ```
 Lets update the captains name:
 ```jldoctest spaceship
-julia> s.captain.name = "JULIA"
+julia> s.captain.name = :JULIA
```
tkf: So Julia 1.0 doesn't do the automatic conversion. I have to change it everywhere.

jw3126:
Should this be merged, or is there something left to do?

tkf:
Yes, it's ready for merge.

PR #44, opened 2018-08-19
Multiproperty

jw3126:


test/runtests.jl
```diff
@@ -1,20 +1,8 @@
...
-    if VERSION < v"0.7-"
-        return esc(ex)
-    else
-        return esc(:(Test.@test_deprecated $ex))
-    end
-end
-
+using Test
 using Setfield
+const P = Setfield
```
tkf: Do you use it somewhere?

test/test_core.jl
```diff
@@ -218,5 +248,6 @@ end
     l = @lens _.x
     @test get(l,t) === 1
 
+    # do we want this to throw an error?
     @test_throws ArgumentError (@set t.z = 3)
```
tkf: I think it could be useful in some situations but you can already use `(t..., new_field=1)` for a named tuple `t`.  Maybe when adding a field to the deeply nested named tuple?  But then there is `modify`. It also breaks (?) the lens law `set(lens, obj, get(lens, obj)) == obj`.

If you don't need it immediately, `throw`ing for now sounds like a good choice to me.

tkf:
LGTM!

test/test_core.jl
jw3126: I had exactly the same thought (might be useful but can be done with `(t..., new_field=1)`.

PR #45, opened 2018-08-20
Perf

jw3126:


test/perf.jl
```diff
@@ -0,0 +1,82 @@
...
+                             (lens_set_a, hand_set_a),
+                             (lens_set_ab, hand_set_ab),
+                             (lens_set_a_and_b, hand_set_a_and_b)
+                            ]
+        b_lens = @benchmark $f_lens($obj, $val)
+        b_hand = @benchmark $f_hand($obj, $val)
+
+        # actually they should be equal
+        # but there is too much noise
+        @test minimum(b_lens).time < 2*minimum(b_hand).time
```
tkf: Maybe you've already tried, but I think another way to do this to use `judge` and make the test pass if the lens is either marked as faster or invariant.  IIUC `judge` would take the fluctuations in the samples into account so it could be more robust.

tkf:
I really like your attitude toward performance!

If benchmarking directly causes too much noise, maybe look at LLVM IR like ForwardDiff does with SIMDTest?:
https://github.com/JuliaDiff/ForwardDiff.jl/blob/655d28502c7c2e0e19b01cc8c921e5712d0797e6/test/SIMDTest.jl#L21
https://github.com/JuliaDiff/ForwardDiff.jl/blob/655d28502c7c2e0e19b01cc8c921e5712d0797e6/.travis.yml#L15

test/perf.jl
jw3126: Thats a good idea. I will try it. Also need to study `BenchmarkTools` more closly, there might be other useful functionality.

jw3126:
In ForwardDiff case, the relevant optimization is SIMD and that happens at LLVM level. AFAIK the -o3 flag is only used by LLVM optimizer, not by the julia optimizer. 
In our case the things that need to optimized are:
* No constructions of lens objects
* Everything should be inlined
* No constructions of intermediate objects that are not contained in the output. E.g. in
```julia
function lens_set_a_and_b(obj, val)
    o1 = @set obj.a = val
    o2 = @set o1.b = val
end
```
o1 should not be constructed. These optimizations happen in julia IR I think. (LLVM might in principle also be able to optimize indermediate objects away. But the julia optimizer seems better at this and if it is not able to optimize an object out,  think LLVM will not be able to either.)
So I think all these can be tested on `@code_typed` as follows:
* Inlining: There are no `:invoke` expressions
* Intermediate objects/lenses: There the same number of `:new` expressions in the lens and hand tunded version.

test/perf.jl
```diff
@@ -0,0 +1,91 @@
...
+function lens_set_a_and_b(obj, val)
+    o1 = @set obj.a = val
+    o2 = @set o1.b = val
+end
+
+function hand_set_a_and_b(obj, val)
+    AB(val, val)
+end
+
+function uniquecounts(iter)
```
jw3126: This helper counts how often an element occurs in a collection. 
```julia
julia> uniquecounts([1,2,1,3,2,2,2])
Dict{Int64,Int64} with 3 entries:
  2 => 4
  3 => 1
  1 => 2
```
I feel like reinventing the whell here, @tkf do you know if such functionality is somewhere in Base or stdlib?

tkf: I don't know.  But maybe not, as there is `countmap` in non-std package `StatsBase` http://juliastats.github.io/StatsBase.jl/latest/counts.html#StatsBase.countmap.

tkf:
That's too bad that the benchmark is hard even for `judge`.  But the factor of 2 may just be good enough.  Let's keep "sampling" the test results in Travis.

Counting instructions from `@code_typed` sounds like a very good idea. That's hopefully robust enough against changes in our code and also in `julia` compiler.

jw3126:
I thought about your words, that we should look at benchmarks from time to time. And then I realized, that this package already had benchmarks and I never really looked at them...
So I thought maybe a more elaborate comparison is reasonable. I am still a bit torn about this. I usually prefer dirt simple test code. What do you think?

jw3126:
I like the IR test, I don't think its overly complex and I am pretty sure it will catch any codegen regression. If it turns out to be maintainence heavy because of upstream IR/codgen changes, we can still replace/remove it.

Accumulating benchmarks over time would be super cool. But I also think its very tricky. I think travis will give you wildly varying performance on each CI run. Running against baseline seems feasible, but still non trivial.
In `PkgBenchmark` there is some [hints](https://juliaci.github.io/PkgBenchmark.jl/stable/export_markdown.html) how to get a nice report. However `PkgBenchmark` is for testing against status quo, while here we have the luxury, that 
* we know the best possible hand tuned implementation
* we can reach this performance

So I would really like to compare these two.

(How) do you test performance regressions over at `DiffEq`?

tkf:
I don't think there are any benchmarks in DiffEq CI.  I know there is an extensive set of benchmarks in https://github.com/JuliaDiffEq/DiffEqBenchmarks.jl but I guess they are run in Chris's machine.  Julia itself has https://github.com/JuliaCI/Nanosoldier.jl which stores benchmarks reports in https://github.com/JuliaCI/BaseBenchmarkReports. That's the best example I know (though I think it requires you to run a bot on a HPC cluster).  It's not Julia (sorry) but there is a a Python benchmarking tool called airspeed velocity `asv` (https://asv.readthedocs.io/en/stable/) for tracking benchmark results over time (e.g., used by numpy https://github.com/numpy/numpy/tree/master/benchmarks).  But I think `asv` is more like a post-hoc analysis tool that checks out past revisions and run the benchmarks.  I'm just mentioning it since the example benchmark outputs (https://pv.github.io/numpy-bench/, http://www.astropy.org/astropy-benchmarks/) are the kind of thing I want to see in Julia projects.

jw3126:
`asv` looks really cool. I am leaning towards merging this and if we see it causes trouble remove/fix it.

tkf:
Cool!  Yeah, the code is robust until now so keep experimenting sounds interesting.  I suggest to do something like #46 to help us find the problem when there is some benchmarking issues.

test/test_staticarrays.jl
```diff
@@ -1,11 +1,13 @@
...
-obj = StaticArrays.@SMatrix [1 2; 3 4]
-l = @lens _[2,1]
-@test get(l, obj) == 3
-@test set(l, obj, 5) == StaticArrays.@SMatrix [1 2; 5 4]
-@test setindex(obj, 5, 2, 1) == StaticArrays.@SMatrix [1 2; 5 4]
-
-v = @SVector [1,2,3]
-@test (@set v[1] = 10) === @SVector [10,2,3]
-@test_broken (@set v[1] = π) === @SVector [π,2,3]
+let
```
tkf: FYI I usually wrap each test file in a `module`.  Alternatively, you can wrap the whole test block with `@testset begin ... end`.  There is some discussion about this here https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/319

tkf: I guess I wasn't clear.  If you put tests in `@testset begin ... end` *inside* `test/test_staticarrays.jl` file then it creates a namespace.  But `@testset begin include("test/test_staticarrays.jl") end` doesn't, as you observed.

tkf:
Oops. There was a benchmark test failure... :sweat_smile:  https://travis-ci.org/jw3126/Setfield.jl/jobs/420508079#L242

test/test_staticarrays.jl
jw3126: I like your idea of wrapping every file in a module. I think I will do that in another PR.

jw3126:
Ouch, lets see if it fails again.

jw3126:
[Here](https://travis-ci.org/jw3126/Setfield.jl/jobs/420528425) is an example of a big fluctuation `TrialEstimate(16.037 ns)` vs `TrialEstimate(11.482 ns)`.

tkf:
Maybe it's just macOS thing, not Julia version?  This is from nightly:

> trials_lens = BenchmarkTools.TrialEstimate[TrialEstimate(28.142 ns), TrialEstimate(27.287 ns), TrialEstimate(28.139 ns), TrialEstimate(28.138 ns), TrialEstimate(27.291 ns), TrialEstimate(28.142 ns), TrialEstimate(27.287 ns), TrialEstimate(29.046 ns), TrialEstimate(29.042 ns), TrialEstimate(28.142 ns)]
> trials_hand = BenchmarkTools.TrialEstimate[TrialEstimate(4.474 ns), TrialEstimate(4.464 ns), TrialEstimate(4.473 ns), TrialEstimate(4.336 ns), TrialEstimate(4.461 ns), TrialEstimate(4.462 ns), TrialEstimate(4.463 ns), TrialEstimate(4.606 ns), TrialEstimate(4.329 ns), TrialEstimate(3.876 ns)]

--- https://travis-ci.org/jw3126/Setfield.jl/jobs/420532092#L243

It looks like the manual implementation is ~7 times faster consistently. Puzzling...

tkf:
Maybe we should construct a benchmark to make sure that `@set` is not getting in the way of optimization by Julia and LLVM (like SIMD) in some more CPU-intensive computations?  In this scenario, time taken by `@set` itself would/could be negligible but surrounding code would be compiled to significantly bad code if we have some performance bug.  The idea is to interleave some getters/setters and some arithmetics, like this:

```julia
@generated function axpy(ls::NTuple{N, Lens}, a, x, y) where N
    quote
        y_0 = y
        Base.Cartesian.@nexprs $N i -> begin
            @inbounds y_i = set(
                ls[i],
                y_{i-1},
                a * get(ls[i], x) + get(ls[i], y_{i-1}))
        end
    end
end
```

This function compiles to LLVM IR (using `-O3`) with SIMD instructions:

```llvm
julia> @ccode_llvm axpy(((@lens _.a), (@lens _.b), (@lens _.c), (@lens _.d)), 1.0,
           (a=1.0, b=2.0, c=3.0, d=4.0),
           (a=1.0, b=2.0, c=3.0, d=4.0))
; Function axpy
; Location: REPL[3]:2
define void @julia_axpy_36458({ double, double, double, double }* noalias nocapture sret, double, { double, double, double, double } addrspace(11)* nocapture nonnull readonly dereferenceable(32), { double, double, double, double } addrspace(11)* nocapture nonnull readonly dereferenceable(32)) {
top:
; Function macro expansion; {
; Location: REPL[3]:5
; Function *; {
; Location: float.jl:399
  %4 = bitcast { double, double, double, double } addrspace(11)* %2 to <4 x double> addrspace(11)*
  %5 = load <4 x double>, <4 x double> addrspace(11)* %4, align 8
  %6 = insertelement <4 x double> undef, double %1, i32 0
  %7 = shufflevector <4 x double> %6, <4 x double> undef, <4 x i32> zeroinitializer
  %8 = fmul <4 x double> %7, %5
;}
; Function +; {
; Location: float.jl:395
  %9 = bitcast { double, double, double, double } addrspace(11)* %3 to <4 x double> addrspace(11)*
  %10 = load <4 x double>, <4 x double> addrspace(11)* %9, align 8
  %11 = fadd <4 x double> %8, %10
;}
  %12 = bitcast { double, double, double, double }* %0 to <4 x double>*
  store <4 x double> %11, <4 x double>* %12, align 8
  ret void
;}
}
```

tkf:
...or generate lens objects in between arithmetics as well. The IR was almost identical.

```julia
@generated function axpy(::Val{fields}, a, x, y) where fields
    ex = quote
        y_0 = y
    end
    for (i, f) in enumerate(fields)
        y_prev = Symbol(:y_, i - 1)
        y = Symbol(:y_, i)
        ex = quote
            $ex
            $y = @set $y_prev.$f += a * x.$f
        end
    end
    ex
end
```

jw3126:
Whee all tests pass, lets merge quickly :laughing:

jw3126:
I changed it back to a factor of 2. The other thing was too fragile. @tkf are you okay with this?

PR #47, opened 2018-08-25
Run performance tests in a subprocess

tkf:
`Pkg.test` sets `--code-coverage` etc.: https://github.com/JuliaLang/Pkg.jl/blob/d123b48e63b5a5e09cb224c6385a3f512627e4ac/src/Operations.jl#L1311-L1320

So maybe running benchmarks in subprocess helps?


tkf:
It makes it worse! :laughing: https://travis-ci.org/jw3126/Setfield.jl/builds/420544458

PR #51, opened 2018-08-31
Julian signature

### LINKED ISSUE #42: Change call signature to get(obj, lens), set(obj, lens, val) and modify(f, obj, lens)?
Re-reading the docstrings, I remember one gotcha in using Setfield: the order of arguments seems to be inconsistent with the choice in `Base`.

For example, you have `get(collection, key, default)` for associative collections but it's `get(l::Lens, obj)` in Setfield.  Considering that lens is similar to key for the collections, it feels like `get(obj, l::Lens)` would be more natural to Julia users.

Similarly, you have `setproperty!(value, name::Symbol, x)` and `setfield!(value, name::Symbol, x)` but it's `set(l::Lens, obj, val)` in Setfield.  Why don't we use `set(obj, l::Lens, val)`?

Also, `get(f::Function, collection, key)` is kind of similar to `modify` (at least argument types are) so `modify(f, obj, lens)` seems to be the natural choice.

Keeping Setfield natural to Julia users seems to be more important than (say) compatibility with lens in Haskell.  But maybe there are some important reason they chose that particular order and you want to keep it?  (I only know very little about Haskell.)  It would be fun to find out if so.

tkf:
closes #42

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -16,41 +16,50 @@ julia> using Setfield
 
 julia> struct T;a;b; end
 
-julia> t = T("AA", "BB")
+julia> obj = T("AA", "BB")
 T("AA", "BB")
 
-julia> l = @lens _.a
+julia> lens = @lens _.a
 (@lens _.a)
 
-julia> get(l, t)
+julia> get(obj, lens)
 "AA"
 
-julia> set(l, t, 2)
+julia> set(obj, lens, 2)
 T(2, "BB")
 
-julia> t
+julia> obj
 T("AA", "BB")
 
-julia> modify(lowercase, l, t)
+julia> modify(lowercase, obj, lens)
 T("aa", "BB")
 ```
 
 # Interface
 Concrete subtypes of `Lens` have to implement
-* `set(lens, obj, val)`
-* `get(lens, obj)`
+* `set(obj, lens, val)`
+* `get(obj, lens)`
 
 These must be pure functions, that satisfy the three lens laws:
-* `get(lens, set(lens, obj, val)) == val` (You get what you set.)
-* `set(lens, obj, get(lens, obj)) == obj` (Setting what was already there changes nothing.)
-* `set(lens, set(lens, obj, val1), val2) == set(lens, obj, val2)` (The last set wins.)
+
+```jldoctest; output = false, setup = :(using Setfield; obj = (a="A", b="B"); lens = @lens _.a; val = 2; val1 = 10; val2 = 20)
+@assert get(set(obj, lens, val), lens) == val
+        # You get what you set.
+@assert set(obj, lens, get(obj, lens)) == obj
+        # Setting what was already there changes nothing.
+@assert set(set(obj, lens, val1), lens, val2) == set(obj, lens, val2)
+        # The last set wins.
+
+# output
+
+```
 
 See also [`@lens`](@ref), [`set`](@ref), [`get`](@ref), [`modify`](@ref).
 """
 abstract type Lens end
 
 """
-    modify(f, l::Lens, obj)
+    modify(f, obj, l::Lens)
 
 Replace a deeply nested part `x` of `obj` by `f(x)`. See also [`Lens`](@ref).
 """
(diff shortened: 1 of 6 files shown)


src/lens.jl
```diff
@@ -16,78 +16,87 @@ julia> using Setfield
...
-* `get(lens, obj)`
+* `set(obj, lens, val)`
+* `get(obj, lens)`
 
 These must be pure functions, that satisfy the three lens laws:
-* `get(lens, set(lens, obj, val)) == val` (You get what you set.)
-* `set(lens, obj, get(lens, obj)) == obj` (Setting what was already there changes nothing.)
-* `set(lens, set(lens, obj, val1), val2) == set(lens, obj, val2)` (The last set wins.)
+
+```jldoctest; output = false, setup = :(using Setfield; obj = (a="A", b="B"); lens = @lens _.a; val = 2; val1 = 10; val2 = 20)
```
tkf: It is a bit ugly but doctest is the easiest way to make sure I don't make any typos in the sample code...

In REPL, `?Lens` looks fine:

image

(except `# output` part)

HTML rendering is a bit better:

image

src/lens.jl
```diff
@@ -141,31 +150,37 @@ compose(l::Lens) = l
 compose(::IdentityLens, ::IdentityLens) = IdentityLens()
 compose(::IdentityLens, l::Lens) = l
 compose(l::Lens, ::IdentityLens) = l
-compose(l1::Lens, l2 ::Lens) = ComposedLens(l1, l2)
-function compose(ls::Lens...)
+compose(l1::Lens, l2 ::Lens) = ComposedLens(l2, l1)
```
tkf: I didn't touch the order of `ComposedLens`. Should I flip it as well?

src/lens.jl
```diff
@@ -141,31 +150,37 @@ compose(l::Lens) = l
...
     getindex(obj, l.indices...)
 end
-Base.@propagate_inbounds function set(l::IndexLens, obj, val)
+Base.@propagate_inbounds function set(obj, l::IndexLens, val)
     setindex(obj, val, l.indices...)
 end
+
+Base.@deprecate get(lens::Lens, obj)       get(obj, lens)
+Base.@deprecate set(lens::Lens, obj, val)  set(obj, lens, val)
+Base.@deprecate modify(f, lens::Lens, obj) modify(f, obj, lens)
```
tkf: It's helpful unless you want to operate lens using lens :)

src/lens.jl
jw3126: Yes it is confusing if they are reversed. Also its probably better to rename the fields of `ComposedLens` into `outer,inner`. But can be a seperate PR.

tkf:
OK I renamed ComposedLens fields. https://github.com/jw3126/Setfield.jl/pull/51#discussion_r214259940  It should be good to go.

src/lens.jl
```diff
@@ -131,41 +140,69 @@ end
...
+
+julia> get(obj, lens)
+1
+```
+"""
+Base.:∘(l1::Lens, l2::Lens) = compose(l1, l2)
+
+function get(obj, l::ComposedLens)
+    inner_obj = get(obj, l.inner)
+    get(inner_obj, l.outer)
```
jw3126: In composed lens, you named the field `outer`, that should be named `inner` and vice versa. For instance this code is much less confusing, if written like this:
```julia
function get(obj, l::ComposedLens)
    outer_obj = obj  # for clarity
    inner_obj = get(outer_obj, l.outer)
    get(inner_obj, l.inner)
end
```

tkf: Hmm so my interpretation was `inner::Lens` means "lens to get the inner object" but I suppose what you meant was "lens applied to the inner object", right?  But to be honest thought about the second interpretation too.  I started wondering maybe we should call them `.left` and `.right`.

tkf: Ah, never mind, I was somehow thinking left is inner. Left is obviously outer.

tkf:
Fixed https://github.com/jw3126/Setfield.jl/pull/51#discussion_r214499079

src/lens.jl
jw3126: Yes the inner/outer lens acts on the inner/outer object. Another reason is that if you have a function composition `g∘f` then `g` is called outer and `f` is called inner function (probably motivated by the formula `(g∘f)(x) = g(f(x))` where `g` resp. `f` appear outside resp. inside the parenthesis). So why not use the same terminology for composition in other categories.

tkf: We also need the notation `f; g` for those who as confused as I was :)

PR #52, opened 2018-08-31
Make sure show(::MultiPropertyLens) does not throw

tkf:
Something I noticed while #51. To be rebased after #51 is merged. It's just one commit: 1284a47e10c7ef3d1085f6497b1153ab1fbf8191

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -71,6 +71,8 @@ end
             @lens _[1]
             @lens _.a.b[2]
             @lens _
+            MultiPropertyLens((a=@lens(_),))
+            (@lens _.a[1]) ∘ MultiPropertyLens((b = (@lens _[1]),))
         ]
         buf = IOBuffer()
         show(buf, item)
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -63,6 +63,9 @@ end
 end
 
 
+struct UserDefinedLens <: Lens end
+
+
 @testset "show it like you build it " begin
     obj = T(1,2)
     i = 3
@@ -73,6 +76,10 @@ end
             @lens _
             MultiPropertyLens((a=@lens(_),))
             (@lens _.a[1]) ∘ MultiPropertyLens((b = (@lens _[1]),))
+            UserDefinedLens()
+            (@lens _.a) ∘ UserDefinedLens()
+            UserDefinedLens() ∘ (@lens _.b)
+            (@lens _.a) ∘ UserDefinedLens() ∘ (@lens _.b)
         ]
         buf = IOBuffer()
         show(buf, item)
--- a/src/experimental.jl
+++ b/src/experimental.jl
@@ -35,4 +35,10 @@ end
         Expr(:call, :(constructor_of($T)), args...)
     )
 end
+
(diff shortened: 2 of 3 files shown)


jw3126:
I would prefer if all lenses pass this test:
https://github.com/jw3126/Setfield.jl/blob/0f98c9ae060dac34b8b34124028d2a7832923c98/test/test_core.jl#L66

One way to get this is as follows (sorry for the depwarns):
```julia
julia> using Setfield; using Setfield.Experimental

julia> l= MultiPropertyLens((a=(@lens _), b=(@lens _[1])));

julia> Setfield.show_generic(l)
WARNING: importing deprecated binding Base.STDOUT into Setfield.
WARNING: Base.STDOUT is deprecated, use stdout instead.
 in module Setfield
in show_generic at /home/ptwlinuxvm/.julia/dev/Setfield/src/sugar.jl
MultiPropertyLens{NamedTuple{(:a, :b),Tuple{Setfield.IdentityLens,Setfield.IndexLens{Tuple{Int64}}}}}((a = (@lens _), b = (@lens _[1])))

```

tkf:
I guess the tricky part is the nested case:

https://github.com/jw3126/Setfield.jl/blob/0f98c9ae060dac34b8b34124028d2a7832923c98/src/sugar.jl#L135-L144

`Base.show` needs to switch to use `show_generic` by just looking at the "root" lens.  We then need to check if all nested lenses have `print_application` or not and then fallback to `show_generic`.  Or maybe a better formatting would be to use `∘` like this: `(@lens _.c[1]) ∘ MultiPropertyLens((a = (@lens _), b = (@lens _[1])))`. 

Another way is to define a sugar for `MultiPropertyLens` but I think `Base.show` should work even for user-defined lens without `print_application` method. So I prefer the above direction.

jw3126:
You are right, user defined lenses should not throw! Maybe we should do something like this:
```julia
has_atlens_support(l) = false
has_atlens_support(l::Union{IdentityLens, FieldLens...}) = true
has_atlens_support(l::ComposedLens) = has_atlens_support(l.lens1) && has_atlens_support(l.lens2)

function show(io::IO, l::Lens)
    if has_atlens_support(l)
        # show as (@lens ...)
    else
        show_generic(io, lens)
    end
end

function show(io::IO, l::ComposedLens)
      if has_atlens_support(l)
          # show as (@lens ...)
      else
           # show as (...) ∘ (...)
      end
end
```

tkf:
Yes, that sounds good.

test/test_core.jl
```diff
@@ -68,6 +68,8 @@ end
             @lens _[1]
             @lens _.a.b[2]
             @lens _
+            MultiPropertyLens((a=@lens(_),))
+            (@lens _.a[1]) ∘ MultiPropertyLens((b = (@lens _[1]),))
```
jw3126: Can you add the same two tests, with `MultiPropertyLens` replaced by a `struct UserDefinedLens <: Lens`?

tkf: Done!

tkf:
Merged! Perhaps a good timing for releasing 0.3?

Issue #15, opened 2018-02-28
Supporting @lens _[end] etc?


tkf:
Are you planning to support those cases?

```julia
l = @lens _[end]
@test get(l, [1, 2, 3]) == 3

l = @lens _[end-1]
@test get(l, [1, 2, 3]) == 2

l = @lens _[end-1][end]  # some complex case
@test get(l, [[1, 2, 3], []]) == 3

l = @lens _[:end]  # just make sure that symbol index is still ok
@test get(l, Dict(:end => 1)) == 1
```

tkf:
Also with colons, like `@lens _[end-2:end]`.

tkf:
I actually looked at the code and tried to fix it.  But then I realized that you need a little interpreter for the indexing operations involving the arithmetic operations and `:` (and what else?).  You also need to be careful about what to interpret at the macro eval time; e.g., `@lens _[f()-end]` should eval `f()`, replace the result `x` with the original indexing AST, and store `x-end`.  This is not a super hairy problem but requires some non-trivial code.  Is it the only way to do it?

tkf:
I had a quick try: #16. It turned out the indexing lenses lose type stability unless we make the indexing interpreter type stable as well. It means to infer the type of the expression without evaluating it when `@lens` is invoked. I guess it can be done by, e.g., injecting some dummy integer to `end`, eval the index expression, and store the type of the result as the type parameter of `IndexLens`.  But I'm still not sure if it makes the indexing interpreter type stable.

tkf:
Maybe more reasonable approach is to do the eval in `@set` and only allow concrete indices in `IndexLens`.

jw3126:
I would definitely like to support `end` (I think in future there will also be `begin`). However I don't really now how `x[1:end-1]` works internally etc. Does it get lowered to something like `x[1:lastindex(x) - 1]`?

jw3126:
Okay I think I have an idea how to implement this treating the `1:end` mechanism as a blackbox.
```julia
struct OOPLens{S,G,M} <: Lens
    set::S
    get::G
    modify::M
end
```
and 
`@lens _[fancy_index_expression]` expands to
```julia
let 
    set(obj, val) = obj[fancy_index_expression] = val; obj
    ...
   OOPLens(set,get,modify)
end
```
One detail is, that `obj[fancy_index_expression] = val` involves `setindex!` and we might like `setindex`.
We could however patch this by wrapping the object
```julia
[...]

PR #60, opened 2019-01-17
Add ConstIndexLens for type-stable indexing of tuples

### LINKED ISSUE #59: Type-stable indexing of tuples
Indexing tuple with `IndexLens` is currently not type sable:

```julia
julia> @code_warntype get((1, 2.0), @lens _[1])
Body::Union{Float64, Int64}
1 ─ %1 = (Base.getfield)(l, :indices)::Tuple{Int64}
│   %2 = (getfield)(%1, 1)::Int64
│   %3 = (Base.getfield)(obj, %2, $(Expr(:boundscheck)))::Union{Float64, Int64}
└──      return %3
```

How about adding (say) `ConstIndexLens` that stores the index in the type parameter?  I think `@lens _[$1]` would be a good sugar for this.

tkf:
This PR adds a new lens `ConstIndexLens` constructible via `@lens _[$1]` etc.

Not sure if it should go directly to public API.  Is it better to put it in experimental?  Although the sugar `@lens _[$1]` is kind of nice to have (but then it's not possible to hide it in experimental module).

fixes #59

### DIFF
--- a/docs/src/intro.md
+++ b/docs/src/intro.md
@@ -20,7 +20,7 @@ julia> s = SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 ```
 Lets update the captains name:
-```jldoctest spaceship
+```jldoctest spaceship; filter = r" .*$"
 julia> s.captain.name = :JULIA
 ERROR: type Person is immutable
 ```
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -203,6 +203,57 @@ Base.@propagate_inbounds function set(obj, l::IndexLens, val)
     setindex(obj, val, l.indices...)
 end
 
+"""
+    ConstIndexLens{I}
+
+Lens with index stored in type parameter.  This is useful for type-stable
+[`get`](@ref) and [`set`](@ref) operations on tuples and named tuples.
+
+This lens can be constructed by, e.g., `@lens _[\$1]`.  Complex expression
+must be wrapped with `\$(...)` like `@lens _[\$(length(xs))]`.
+
+# Examples
+```jldoctest
+julia> using Setfield
+
+julia> get((1, 2.0), @lens _[\$1])
+1
+
+julia> Base.promote_op(get, typeof.(((1, 2.0), @lens _[\$1]))...)
+Int64
+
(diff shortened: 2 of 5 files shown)


jw3126:
Thanks for looking into this! I am not against this, but I want to ponder the options a bit.

* In principle the compiler should be able to constant propagate this. And in my experience it is very good at optimizing all lenses away. Here is a simple example:
```julia
function doit(tup)
    l = @lens _[1]
    get(tup, l)
end
@code_warntype doit((1, 2.0))
Body::Int64
1 ─ %1 = (Base.getfield)(tup, 1, true)::Int64
└──      return %1
```
Is this type instability a problem you encountered in real code? Can you give such an example?

* One thing that might help the compiler would be to use an approach as sketched [here](
https://github.com/jw3126/Setfield.jl/issues/15#issuecomment-369194231) for `IndexLens`. This would have the benefit, that it also solves `@lens x[end]`. What do you think?

tkf:
As for the "real world" example, I needed this in [`flat` branch](https://github.com/tkf/Transducers.jl/pull/12) of Transducers.jl to gain some type-stability: https://github.com/tkf/Transducers.jl/commit/5ae826a8a9acce1e0a0feb497a6f49da1a252231#diff-61efac4de4204dbab82023b75c3143b5 but I then ditched that branch as I couldn't make it fast.

jw3126:
This is probably more or less your `Val` idea:
```julia
struct StaticLens{l} <: Lens end
static(l::Lens) = StaticLens{l}()
set(o, ::StaticLens{l}, val) where {l} = set(o, l, val)
...
```
And you can use it like:
```julia
static(@lens _[1].bar)
```
But then if you want some static and some dynamic indices this becomes more awkward:
```julia
static(@lens _[i]) ∘ @lens(_[j])
```
What do you think?

jw3126:
Just checked, the sweeper example is correctly optimized with this PR. Still a good test case.

docs/src/intro.md
```diff
@@ -20,7 +20,7 @@ julia> s = SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 SpaceShip(Person(:julia, 2009), [0.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 ```
 Lets update the captains name:
-```jldoctest spaceship
+```jldoctest spaceship; filter = r" .*$"
```
tkf: Not related to this PR, but I fixed this doctest for Julia 1.2 just by ignoring the detail in error message.

jw3126:
Thanks for this PR! High quality as always. Do you want me to tag a new release?

PR #66, opened 2019-05-10
Fix show for IndexLens

tkf:
I realized that `show` for lenses like `@lens _[:a]` and `@lens _["a"]` are not `parse`-able.  This PR fixes it simply by using `repr` when printing indices.

Before:

```julia
julia> @lens _[:a]
(@lens _[a])
```

After:

```julia
julia> @lens _[:a]
(@lens _[:a])
```


jw3126:
LGTM, thanks!

PR #67, opened 2019-05-15
Switch to Project.toml

jw3126:


jw3126:
@tkf do you know why docs fails and whats the correct fix?

tkf:
I think only the first commit 7461f3aede551efbc699f3876023bdf0d1e445d5 is necessary to fix the CI.  But I thought it's a good timing to do a few cleanups (hence the additional few commits).  In particular, I think it should fix the problem that the documentation is not rendered correctly in: https://pkg.julialang.org/docs/Setfield/dVM1R/0.3.4/

jw3126:
Thanks @tkf for doing this!

PR #69, opened 2019-06-22
Use only public APIs in constructor_of

tkf:


jw3126:
Oh nice, I was not aware of `nameof`. LGTM!

PR #81, opened 2019-09-16
Add compat information

tkf:
see https://github.com/JuliaRegistries/General/pull/3509#issuecomment-531686525

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,10 @@ version = "0.4.0"
 [deps]
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 
+[compat]
+MacroTools = "≥ 0.4.4"
+julia = "1"
+
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"


Project.toml
```diff
@@ -5,6 +5,10 @@ version = "0.4.0"
 [deps]
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 
+[compat]
+MacroTools = "≥ 0.4.4"
```
tkf: I copied this from REQUIRE in #67. But compat without upper bound is rather controversial (although many people including myself do so).  The consequence of doing this is rather unclear to me since how General registry and Pkg.jl work is not super transparent.

jw3126: Yeah, I also leave out the upper bound. What do you mean by:

> The consequence of doing this is rather unclear to me since how General registry and Pkg.jl work is not super transparent.

What do you mean by this? I thought the solver tries to use packages with as high version number as possible, with the constraint that the bounds given by all `[compat]` sections of all packages in the environment are satisfied?

jw3126:
LGTM

Project.toml
tkf: Sorry, that was rather my rant...  Basically, once you publish a package without upper bound in a dependency, it is in principle impossible to add upper bound safely. Consider the case MacroTools got to 1.0 and so you want to start adding upper bound say, as of Setfield 0.9.  Then, if MacroTools bumps the major version to 2, Setfield 0.9 is no longer considered compatible but Setfield 0.8 is considered compatible (which likely is not).  But, since JuliaRegistries/General seems to accept manual change in the registry data, this _may_ not be a problem (i.e., we can add the bounds later).

(My preference is to use something like `0.4.4 <= MacroTools < 1`; i.e., being a bit loose until the dependency hit 1.0.  But it's not possible untile https://github.com/JuliaLang/Pkg.jl/pull/1232)

jw3126: I see thanks for explaining!

Issue #68, opened 2019-06-04
constructor_of()


rafaqz:
I'm updating and abstracting Flatten.jl to allow arbitrary flattening and modifications of nested structs using method and type 'queries'.

It has some similarities to Setfield.jl in that it needs to call the constructor of immutable types, so I've just copied your `constructor_of()` to make this more apparent. But I was thinking it might be nice to abstract this to a base package one day so that structs with custom type parameters can just define a specific constructor once for anything that needs it.

jw3126:
I am open to moving things out of `Setfield.jl` that other packages want to share. Are you only thinking about `constructor_of` or do you have other stuff that should move to the base package?
```julia
@generated constructor_of(::Type{T}) where T =
getfield(T.name.module, Symbol(T.name.name))
```
While this function is non trivial, it is also very short and I hesitate to add a dependency just for such a short function.

rafaqz:
Yeah it does seem a bit short to add that dependency for.

I mostly wanted to flag that I'm using the exact same function for the same purpose, and it seems to be a general requirement for any tool that works with immutables like this. Maybe when more work is done in base we can have something generalised.

tkf:
While the default `constructor_of` is super short, I think it makes sense to extract it out to a base package because it is an overloadable interface.  For example, maybe QuickTypes.jl can overload `constructor_of` by default.  We then don't need to use `Setfield.@settable`.  I also occasionally overload `constructor_of` in my code.  It'd be nice if such definition of `constructor_of` can be reused in other packages.

> arbitrary flattening and modifications of nested structs

Just FYI I recently created a similar package https://github.com/tkf/Kaleido.jl based on Setfield.jl

rafaqz:
Yeah I was thinking that, but I'm also not strongly opinionated about it because in practice I only have one or two types that I'm actually overloading `constructor_of` for. I imagine some coding styles that use a lot of custom type parameters would need it a lot more. But I would depend on ConstructorBase (or whatever) in Flatten if we split it out.

I'll take a look at Kaleido. The core difference between Flatten and Setfield/Kaleido approach is that Flatten is more about abstracting the process than specifying it - for cases where you don't know or care what the field names are. Its like a query: get all the e.g `Real` fields in some user-defined nested struct, do something to them and reconstruct the object. It mostly compiles down to just modifying those fields directly.

tkf:
[...]

PR #84, opened 2019-09-25
Define constructor_of for singleton lenses

tkf:
While I'm using `constructor_of` inside [some meta-programming code](https://github.com/tkf/ChainCutters.jl/pull/8/files), I noticed that it didn't work with some lenses.  It's not necessary for Setfield.jl to work so I'm not sure if this is a good change.  But it'd make sense for Setfield.jl to have its own type `constructor_of`-compatible?

A more magic way to do this is

```julia
@generated constructor_of(::Type{T}) where T =
    if fieldcount(T) == 0
        T
    else
        getfield(parentmodule(T), nameof(T))
    end
```

Not sure if I like it though...

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -324,6 +324,13 @@ FunctionLens(f) = FunctionLens{f}()
 
 get(obj, ::FunctionLens{f}) where f = f(obj)
 
+constructor_of(::Type{T}) where T <: Union{
+    # List of types that should/can not change type parameters:
+    PropertyLens,
+    ConstIndexLens,
+    FunctionLens,
+} = T
+
 Base.@deprecate get(lens::Lens, obj)       get(obj, lens)
 Base.@deprecate set(lens::Lens, obj, val)  set(obj, lens, val)
 Base.@deprecate modify(f, lens::Lens, obj) modify(f, obj, lens)
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -337,6 +337,12 @@ end
     @test obj2 === TT(2, :three)
 end
 
+@testset "types that do not change parameter via constructor_of" begin
+    @test constructor_of(typeof(@lens _.a))() === @lens _.a
+    @test constructor_of(typeof(@lens _[$1]))() === @lens _[$1]
+    @test constructor_of(typeof(@lens eltype(_)))() === @lens eltype(_)
+end
+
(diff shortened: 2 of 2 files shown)


jw3126:
Any trouble you anticipate with the more magic implementation?

tkf:
I find it uneasy that the behavior is very different when `fieldcount(T) == 0`.

Actually, the magic way can be implemented by the user side.  That is to say, you can simple do

```julia
@generated my_constructor_of(::Type{T}) where T =
    if fieldcount(T) == 0
        T
    else
        :(Setfield.constructor_of(T))
    end
```

So, I think it is better to keep `constructor_of` in `Setfield` as simple as possible.

jw3126:
Okay, I guess then the question is what the semantics of `constructor_of` should be. I thought 
```julia
T = typeof(x)
constructor_of(T)(fieldvalues(x)...) == x
```
is a property that we would like to have.

Your interpretation seems to be that `constructor_of` just strips all type parameters? That is a useful function as well, but maybe should have a different name?

tkf:
> Your interpretation seems to be that `constructor_of` just strips all type parameters? That is a useful function as well, but maybe should have a different name?

We can always do `typeof(x)(fieldvalues(x)...)`.  So, if you don't need to change type parameters, I think that's the way to go.

> then the question is what the semantics of `constructor_of` should be.

I'd say "`constructor_of(typeof(T))(fieldvalues(x)...) == x` holds _and_ `constructor_of(T)(args...)` works as many types of `args` as possible" (though I guess that's somewhat vague).  The problem is that in some cases type parameters cannot be determined from the arguments and you need to define `constructor_of` for your types.  So, I thought it'd be better to keep the default implementation simple so that it is clear when you need to define it.

jw3126:
>I'd say "constructor_of(typeof(T))(fieldvalues(x)...) == x holds and constructor_of(T)(args...) works as many types of args as possible" (though I guess that's somewhat vague). 

I like that specification.

>So, I thought it'd be better to keep the default implementation simple so that it is clear when you need to define it.

Thats a sane approach, I am ok with it. If we are really pedantic about it, we should merge this PR. But I think its okay to only merge things like this PR if there is practical need.

jw3126:
BTW I think we should we create a `ConstructionBase` package now like discussed in https://github.com/jw3126/Setfield.jl/issues/68
I think you would be a great owner for that package. If you don't want to, I can create it instead.

PR #86, opened 2019-09-25
Dispatch to text/plain show if defined in downstreams

tkf:
Context: I'm playing with pretty-printer for Kaleido.jl https://github.com/tkf/Kaleido.jl/pull/6 and noticed that there is no easy way to extend `show(io::IO, ::MIME"text/plain", l::Lens)`. For example, if I call `show(stdout, "text/plain", mylens ∘ (@lens _.a))`, Julia calls `show(::IO, ::MIME"text/plain", ::Any)` which calls `show(::IO, ::ComposedLens)`. So, by the time my `show` method is called, the mime type is gone.

I think the easiest would be to pass around the mime type in Setfield.jl.

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -201,21 +201,29 @@ function print_application(printer, io, l::ComposedLens)
     end
 end
 
-function Base.show(io::IO, l::Lens)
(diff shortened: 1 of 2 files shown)


jw3126:
LGTM!

PR #88, opened 2019-09-28
Fix text/plain show to avoid StackOverflowError

tkf:
PR #86 introduced a horrible bug that causes `StackOverflowError` when a user-defined function does not define text/plain `show`.  This PR fixes it by manually detecting recursion.  It's rather ugly so I'm open to other solutions (e.g., rename `_show` to `showlens` and expose it as a public API).

### DIFF
--- a/src/Setfield.jl
+++ b/src/Setfield.jl
@@ -8,4 +8,12 @@ include("sugar.jl")
 include("functionlenses.jl")
 include("settable.jl")
 include("experimental.jl")
+
+for n in names(Setfield, all=true)
+    T = getproperty(Setfield, n)
+    if T isa Type && T <: Lens && (T === ComposedLens || has_atlens_support(T))
+        @eval Base.show(io::IO, l::$T) = _show(io, nothing, l)
+    end
+end
+
 end
--- a/src/Setfield.jl
+++ b/src/Setfield.jl
@@ -9,6 +9,13 @@ include("functionlenses.jl")
 include("settable.jl")
 include("experimental.jl")
 
+# To correctly dispatch to `show(::IO, ::CustomLens)` when it is defined by a
+# user, we avoid defining the generic `show(::IO, ::Lens)`.  This way, we can
+# safely call `show` inside `ComposedLens` without worrying about the
+# `StackOverflowError` that can be easily triggered in the previous approach.
+# See also:
+# * https://github.com/jw3126/Setfield.jl/pull/86
+# * https://github.com/jw3126/Setfield.jl/pull/88
 for n in names(Setfield, all=true)
     T = getproperty(Setfield, n)
     if T isa Type && T <: Lens && (T === ComposedLens || has_atlens_support(T))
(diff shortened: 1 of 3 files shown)


jw3126:
If https://github.com/JuliaRegistries/General/pull/3869 can be stopped, we can think about a better solution. Your idea of exposing `showlens` would it force the user to implement it for new lenses? I think to define a new lens only `set` and `get` should be mandatory and nothing else.

tkf:
> only `set` and `get` should be mandatory

I agree. The idea is that `showlens` would fallback to `show_generic` so that Setfield API is not changed.  It's just that users (maybe just me?) who want to modify printing behavior has to overload `showlens` instead of `show`.

jw3126:
Ok so only users who want to overload `show` and need to dispatch on `mime` would need `showlens`? Overloading `show(io::IO, lens::MyLens)` would still work right?

tkf:
> Overloading `show(io::IO, lens::MyLens)` would still work right?

In the `showlens` approach, it would work only for directly `show`ing `MyLens`. If `MyLens` is inside a `ComposedLens`, it would be printed via the default `showlens` which calls `show_generic`.

jw3126:
Ah right. What if we do not define `show(io::IO, [mime,] l::Lens)`, but instead
`show(io::IO, [mime,], l::Union{all lenses in Setfield})`.
Would that prevent the stack overflow?

tkf:
That's a good idea. I implemented it here: de7e8e7ac711e007530453a2a82477e78fc75d60

src/Setfield.jl
```diff
@@ -8,4 +8,12 @@ include("sugar.jl")
 include("functionlenses.jl")
 include("settable.jl")
 include("experimental.jl")
+
+for n in names(Setfield, all=true)
+    T = getproperty(Setfield, n)
+    if T isa Type && T <: Lens && (T === ComposedLens || has_atlens_support(T))
+        @eval Base.show(io::IO, l::$T) = _show(io, nothing, l)
+    end
+end
```
jw3126: Can you add a comment that mentions the stack overflow and this PR?

tkf: See bd1594aedf9a6ef68779c6fb537cf4887cafc61b

src/sugar.jl
```diff
@@ -234,10 +236,3 @@ function print_in_atlens(io, l)
     end
     print(io, ')')
 end
-
-function show_generic(io::IO, args...)
```
tkf: A nice side-effect was that we can get rid of `show_generic` now.

jw3126: Yes that is great!

test/test_core.jl
```diff
@@ -356,13 +356,6 @@ Setfield.constructor_of(::Type{<: B{T}}) where T = B{T}
...
-@testset "show_generic" begin
-    l = @lens _[1]
-    s = sprint(Setfield.show_generic,l)
-    l2 = eval(Meta.parse(s))
-    @test l == l2
-end
-
 @testset "text/plain show" begin
     @testset for lens in [
         LensWithTextPlain()
```
jw3126: Maybe also compose this lens?

jw3126: nvm did not see it was already there.

jw3126:
I think we can merge this. Before tagging the release can you maybe work with this version for a week and see if other issues come up?

tkf:
> Before tagging the release can you maybe work with this version for a week and see if other issues come up?

Yup. I'll do reverse-dependency testing with Kaleido and its downstreams. (That's exactly how I found this bug.)

PR #91, opened 2019-10-01
use ConstructionBase

jw3126:


Project.toml
```diff
@@ -3,6 +3,7 @@ uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 version = "0.4.1"
```
tkf:
```suggestion
 version = "0.5.0"
```

This is backward incompatible so we need to bump the leading version.

Project.toml
```diff
@@ -1,8 +1,9 @@
 name = "Setfield"
 uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-version = "0.4.1"
+ version = "0.5.0"
```
tkf:
```suggestion
version = "0.5.0"
```

Hmm... It looks like GitHub introduces a leading space when I use the suggestion feature?

test/test_quicktypes.jl
```diff
@@ -128,7 +128,7 @@ end
 # Another way to "support" QuickTypes with type parameters is to use
 # QuickTypes.construct.
 @qstruct_fp Plane2(nwheels, weight::Number; brand=:zoomba)
-Setfield.constructor_of(::Type{<: Plane2}) =
+Setfield.constructorof(::Type{<: Plane2}) =
```
tkf: I suppose we need to use `ConstructionBase.constructorof` here?  Same for `test_settable.jl`.

jw3126:
@tkf Do you know why CI fails? Locally I can `]add ConstructionBase` just fine. Does Travis some magic caching?

tkf:
I think it's working https://travis-ci.org/jw3126/Setfield.jl/builds/595455069 (click `continuous-integration/travis-ci/pr` above the merge button)? If you are looking at `continuous-integration/travis-ci/push`, that's the CI for this branch as-is built when you pushed, different from the one you can restart via close-then-open.

jw3126:
Ah I see, thanks for explaining.

tkf:
image

You can also click this right-top `Restart build` button.

jw3126:
@tkf I would like to tag a release, any objections?

tkf:
I think it's good to go. FYI, testing Kaleido with Setfield 0.5 worked: https://github.com/tkf/Kaleido.jl/pull/8

jw3126:
@JuliaRegistrator register

PR #94, opened 2019-10-10
Link projects using Setfield

tkf:
VegaLite.jl started using Setfield as property modification API https://github.com/queryverse/VegaLite.jl/pull/191. I think it makes sense to link VegaLite as it uses Setfield in a non-trivial way. I also smuggled a link to Kaleido as well :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,3 +14,11 @@ using Setfield
 @set obj.a.b.c = d
 ```
 For more information, see [the documentation](https://jw3126.github.io/Setfield.jl/latest/intro/).
+
+# Projects using Setfield
+
+* [VegaLite.jl](https://github.com/queryverse/VegaLite.jl) overloads
+  `getproperty` and lens API to manipulate JSON-based nested objects.
+
+* [Kaleido.jl](https://github.com/tkf/Kaleido.jl) is a library of
+  additional lenses.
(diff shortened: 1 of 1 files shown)


jw3126:
With `Pkg2`, there was a function that could list all downstream packages, but it seems that it is [gone](https://github.com/JuliaLang/Pkg.jl/issues/871).

jw3126:
[PhaseSpaceIO](https://github.com/jw3126/PhaseSpaceIO.jl) has some fancy [`setproperties` overloading](https://github.com/jw3126/PhaseSpaceIO.jl/blob/v0.2.0/src/egs/particle.jl#L3). Though I guess I am the only user of that package.

tkf:
Wow, properties packed in UInt? That's crazy :)

I don't think we need to list all packages just using Setfield.jl (that's probably a long and not so interesting list).  My aim was listing some creative use of Setfield so that it can inspire crazier use cases.  So I think it makes sense to add PhaseSpaceIO there.

jw3126:
> 
> Wow, properties packed in UInt? That's crazy :)
> 
True, but not as crazy as the code you need to manipulate these without `setproperties` :).

jw3126:
I think an example, where `@lens` is used for differentiation would be nice. Do you have one on github?

tkf:
I have https://github.com/tkf/Bifurcations.jl and https://github.com/tkf/SteadyStateFit.jl but they are not registered.  For interop with auto-diff systems, I think it makes sense to add it as a tutorial/demo in Kaleido.jl (and add a link from Setfield's README) since that's my main use case of Kaleido.jl.

PR #95, opened 2019-10-12
Lenstransform

jw3126:
Implementing @tkf gist https://github.com/jw3126/Setfield.jl/pull/92#issuecomment-540434449


src/sugar.jl
```diff
@@ -133,7 +133,7 @@ struct _UpdateOp{OP,V}
 end
 (u::_UpdateOp)(x) = u.op(x, u.val)
 
-function atset_impl(ex::Expr; overwrite::Bool)
+function atset_impl(ex::Expr; overwrite::Bool=false, lenstransform=identity)
```
jw3126: Let's rename these functions. How about `lensmacro`/`setmacro`? Also a docstring should be added.

test/test_custom_macros.jl
```diff
@@ -0,0 +1,78 @@
+module CustomMacros
```
jw3126: I think this could be a `Literate` page in the docs.

src/sugar.jl
```diff
@@ -142,14 +142,15 @@ function atset_impl(ex::Expr; overwrite::Bool)
...
     ret = if ex.head == :(=)
         quote
-            lens = $lens
+            lens = ($lenstransform)($lens)
             $dst = set($obj, lens, $val)
         end
     else
         op = get_update_op(ex.head)
-        f = :(_UpdateOp($op,$val))
+        f = :($_UpdateOp($op,$val))
```
tkf: I think changes like this need to happen in `parse_obj_lenses` as well. For example, `lens = :($IndexLens($index))`.

src/sugar.jl
```diff
@@ -188,12 +189,16 @@ julia> set(t, (@lens _[1]), "1")
 
 """
 macro lens(ex)
+    atlens_impl(ex)
```
tkf: I guess we don't need to expose `atlens_impl` as people can just do

```julia
macro mylens(ex)
    :(mylenstransform(@lens $ex))
end
```

? Or maybe with some more `$` and `esc`. :)

test/test_custom_macros.jl
tkf: That's a great idea!

We also should define a macro in an isolated sub-module and import it into the namespace we use for testing. This is for catching some possible namespacing issues.

src/sugar.jl
tkf: `setmacro` sounds good.

src/sugar.jl
jw3126: It is nice, that this is a one liner. On the other hand the fact that you are asking about `$` and `esc` shows that it is somewhat non trivial to get isolation right. So I left `lensmacro` alive.

examples/custom_macros.jl
```diff
@@ -0,0 +1,75 @@
...
+@test o.b == 20
+
+l = Lens!(@lens _.foo[1])
+o = (foo=[1,2,3], bar=:bar)
+set(o, l, 100)
+@test o == (foo=[100,2,3], bar=:bar)
+
+# Now we can implement the syntax macros
+
+using Setfield: setmacro, lensmacro
```
tkf:
```suggestion
module Clone
using Setfield: setmacro, lensmacro
```

examples/custom_macros.jl
```diff
@@ -0,0 +1,75 @@
...
+
+macro myset!(ex)
+    setmacro(Lens!, ex)
+end
+
+macro mylens!(ex)
+    lensmacro(Lens!, ex)
+end
+
```
tkf:
```suggestion

end#module Clone

using .Clone: Clone
```

Issue #85, opened 2019-09-25
Implement show for typeof(@lens myfunction(_))


tkf:
Re: https://github.com/jw3126/Setfield.jl/pull/77#issuecomment-530664735

> > > * Might also make sense to actually print this type as `typeof(@lens myfunction(_))`?
> > 
> > 
> > We can. But we need to be _extremely_ careful in the implementation: [JuliaLang/julia#29428](https://github.com/JuliaLang/julia/issues/29428)
> 
> Oh wow I somehow never ran into that bug.

Actually, I noticed that the Julia "bug" has been fixed since Julia 1.1.0. So I think it's (relatively) safe to implement this.

(Just recording it as a possible improvement.)

PR #97, opened 2019-10-20
better printing for lens types

simeonschaub:
Addresses #85. Also uses this better printing for types `PropertyLens`, `ConstIndexLens` and `ComposedLens`.

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -303,3 +303,21 @@ function print_in_atlens(io, l)
     end
     print(io, ')')
 end
+
+_instance(T::Type{<:Union{PropertyLens,ConstIndexLens,FunctionLens}}) = T()
+function _instance(T::Type{ComposedLens{U,V}}) where {U,V}
+    instances = _instance.((U, V))
+    any(==(nothing), instances) && return nothing
+    return T(instances...)
+end
+_instance(::Any) = nothing
+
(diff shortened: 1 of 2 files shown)


jw3126:
Thanks for this and #98. The main motivation for changing `show` of `FunctionLens` was, that the type parameters of `FunctionLens` are not part of the stable API. See https://jw3126.github.io/Setfield.jl/latest/#Setfield.FunctionLens and https://github.com/jw3126/Setfield.jl/pull/77#issuecomment-530551511 . So it makes sense to print this type like we want people to use it in overloading signatures. That does not apply to the other lenses. So I am unsure if I like "hiding" their type. What do you think @tkf?

tkf:
@simeonschaub Thanks for giving this a shot!

@jw3126 Yeah, doing only for `FunctionLens` makes sense.  Overloading `Type{...}` still is not really the safest thing we can do.  I know Zygote.jl does it but it also was bitten by it https://github.com/FluxML/Zygote.jl/issues/134

Also, IIRC `show` for lens values do not differentiate `a ∘ (b ∘ c)` and `(a ∘ b) ∘ c`.  Doing this at type level would be confusing.

simeonschaub:
Ok, thanks for the feedback! Closing this in favor of #99

PR #99, opened 2019-10-21
custom type printing only for FunctionLens

### LINKED ISSUE #85: Implement show for typeof(@lens myfunction(_))

simeonschaub:
More minimal successor to #97. Close #85.

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -170,7 +170,7 @@ end
     buf = IOBuffer()
     flens = @lens first(_)
     show(buf, typeof(flens))
-    @test String(take!(buf)) == "typeof$flens"
+    @test String(take!(buf)) == "typeof(@lens first(_))"
 
     # test correct printing of UnionAll
     show(buf, Setfield.FunctionLens)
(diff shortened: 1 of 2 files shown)


test/test_core.jl
```diff
@@ -158,9 +158,25 @@ Base.show(io::IO, ::MIME"text/plain", ::LensWithTextPlain) =
...
+        typeof_item2 = eval(Meta.parse(String(take!(buf))))
+        @test typeof(item) === typeof_item2
     end
 end
 
+@testset "show of typeof(::FunctionLens)" begin
+    buf = IOBuffer()
+    flens = @lens first(_)
+    show(buf, typeof(flens))
+    @test String(take!(buf)) == "typeof$flens"
```
jw3126: Lets be completly explicit:
```suggestion
    @test String(take!(buf)) == "typeof(@lens first(_))"
```

jw3126:
Thanks again for your PRs

PR #102, opened 2019-11-05
Upper-bound dependencies

tkf:
I think now it's better to upper bound so that registration would be automated.

FYI it looks like we can use https://github.com/bcbi/CompatHelper.jl to auto-bump compat via GitHub action.


jw3126:
Okay. I would prefer to do it manually for until `CompatHelper` has an official release.

tkf:
IIUC CompatHelper.jl just sends a PR with updates in Project.toml and you still have to click the merge button manually.  So, the worst-case scenario seems to be that you just waste some CI time.

PR #103, opened 2019-11-15
Bump ConstructionBase compat bound

tkf:


jw3126:
Can we bump to `1` instead of `1.0`? My understanding is, that the former will allow for e.g. `ConstructionBase 1.2.6`, but the latter will not.

tkf:
`1` and `1.0` for compat are actually the same, since version `1.y` is compatible to `1.x` for any `x`, `y` s.t. `y >= x`:

```julia
julia> Pkg.Types.semver_spec("1")
VersionSpec("1")

julia> Pkg.Types.semver_spec("1.0")
VersionSpec("1")
```

Pre-1.0 versions behave differently, though.

tkf:
But let's use `ConstructionBase = "0.1, 1"` anyway to be consistent with `julia = "1"`.

jw3126:
I set up `CompatHelper.jl`, let's see if it works.

jw3126:
If this PR is urgent for you feel free to merge. Otherwise I wouldd prefer to watch compat helper doing it.

tkf:
:+1: for CompatHelper.jl. This PR is not urgent. I was just updating other packages and realized that Setfield needs this too. I think it'll be a good test case for CompatHelper.jl setup.

jw3126:
Indeed. You have any experience, how long CompatHelper needs to open a PR?

tkf:
It's determined by

https://github.com/jw3126/Setfield.jl/blob/bffd20d93f406f3b9677cece719ba7509e68019c/.github/workflows/CompatHelper.yml#L3-L5

So, once per hour.

If you want to experiment with it, you can add `types: [opened, reopened]` and open/close an issue to trigger the action:

```yaml
on:
  schedule:
    - cron: '00 * * * *'
  issues:
    types: [opened, reopened]
```

But (I think) this requires you to tweak the yaml file in `master` branch which is not really great.  To try things with github actions, I do it in a dummy repository https://github.com/tkf/tmp (i.e., force-push another repository to the `master`).

PR #114, opened 2019-12-16
Remove show(::IO, ::Type{<:FunctionLens{S}}) (revert #99)

### LINKED ISSUE #113: Remove type piracy for Base.show
[This](https://github.com/jw3126/Setfield.jl/blob/v0.5.3/src/sugar.jl#L311) method for `Base.show` causes all sorts of trouble. For example, it breaks `methods(Base.show)`. It also takes down the symbol server julia process for the VS Code extension with a segfault...

I think at the end of the day this is a particularly bad case of type piracy because it really breaks julia base. Could you remove it sooner rather than later and tag a new release?

tkf:
It turned out this method causes segfaults in VS Code extension. I think we better play on the safe side.

close #113

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -307,8 +307,3 @@ function print_in_atlens(io, l)
     end
     print(io, ')')
 end
-
-function Base.show(io::IO, T::Type{<:FunctionLens{S}}) where S
-   print(io, "typeof")
-   show(io, T())
-end
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -170,7 +170,7 @@ end
     buf = IOBuffer()
     flens = @lens first(_)
     show(buf, typeof(flens))
-    @test String(take!(buf)) == "typeof(@lens first(_))"
+    @test String(take!(buf)) == "typeof$flens"
 
     # test correct printing of UnionAll
     show(buf, Setfield.FunctionLens)
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -158,23 +158,7 @@ Base.show(io::IO, ::MIME"text/plain", ::LensWithTextPlain) =
         show(buf, item)
         item2 = eval(Meta.parse(String(take!(buf))))
         @test item === item2
-
-        # showing of Type{<:Lens}
-        show(buf, typeof(item))
-        typeof_item2 = eval(Meta.parse(String(take!(buf))))
-        @test typeof(item) === typeof_item2
     end
-end
-
-@testset "show of typeof(::FunctionLens)" begin
-    buf = IOBuffer()
-    flens = @lens first(_)
-    show(buf, typeof(flens))
-    @test String(take!(buf)) == "typeof$flens"
-
-    # test correct printing of UnionAll
-    show(buf, Setfield.FunctionLens)
-    @test String(take!(buf)) == "Setfield.FunctionLens"
 end
 
 function test_getset_laws(lens, obj, val1, val2)


tkf:
@jw3126 How about releasing a new version just after merging this?

davidanthoff:
Actually, I have a new theory that this might not be the root cause. Maybe we wait with this PR until I have investigated my new theory?

tkf:
Sounds good to me.

BTW, maybe something like this is happening in the VS code extension:

```julia
julia> show(stdout, Base.unwrap_unionall(Setfield.FunctionLens))
typeofERROR: UndefVarError: f not defined
Stacktrace:
 [1] (::Type{typeof
SYSTEM (REPL): showing an error caused an error
ERROR: UndefVarError: f not defined
Stacktrace:
 [1] (::Type{typeof
SYSTEM (REPL): caught exception of type UndefVarError while trying to handle a nested exception; giving up
```

A MWE is

```julia
julia> struct MyType{f} end

julia> function demo(T::Type{<:MyType})
           println("Calling T()...")
           y = T()
           println("DONE")
           return y
       end
demo (generic function with 1 method)

julia> demo(Base.unwrap_unionall(MyType))
Calling T()...
ERROR: UndefVarError: f not defined
Stacktrace:
 [1] (::Type{MyType{f}})() at ./REPL[1]:1
 [2] demo(::Type{MyType{f}}) at ./REPL[2]:3
 [3] top-level scope at REPL[3]:1
```

tkf:
We can "fix" this by checking `@isdefined(S)`:

```julia
julia> function demo2(T::Type{<:MyType{S}}) where S
           @isdefined(S) || return :abort
           println("Calling T()...")
           y = T()
           println("DONE")
           return y
       end
demo2 (generic function with 1 method)

julia> demo2(Base.unwrap_unionall(MyType))
:abort
```

So a better definition of the `show` is:

```diff
diff --git a/src/sugar.jl b/src/sugar.jl
index 94cb6be..cb308df 100644
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -309,6 +309,7 @@ function print_in_atlens(io, l)
 end

 function Base.show(io::IO, T::Type{<:FunctionLens{S}}) where S
+   @isdefined(S) || return invoke(show, Tuple{IO,Type}, io, T)
    print(io, "typeof")
    show(io, T())
 end
```

davidanthoff:
For now I added https://github.com/julia-vscode/SymbolServer.jl/pull/65, which seems to work around the problem entirely. I don't really understand what the root cause is :) But I suspect that the `show` problem was only a symptom, i.e. there was an error, and then when it tried to show that error message it somehow interacted badly with the `show` method. So probably your MWE here gets closer to what is happening.

tkf:
Yeah, so that's the same error (my "fix" above fixes it):

```julia
julia> Base.arg_decl_parts(@which show(stdout, typeof(@lens identity(_))))
ERROR: UndefVarError: f not defined
Stacktrace:
 [1] (::Type{typeof
SYSTEM (REPL): showing an error caused an error
ERROR: UndefVarError: f not defined
Stacktrace:
 [1] (::Type{typeof
SYSTEM (REPL): caught exception of type UndefVarError while trying to handle a nested exception; giving up
```

Grepping `while trying to handle a nested exception`, I noticed that this error handling was REPL-specific:

https://github.com/JuliaLang/julia/blob/a0b67dc0c08afe7ccb268057cc87b0d5405f4a63/stdlib/REPL/src/REPL.jl#L173-L186

Logging has another fix. Running

```sh
julia --startup-file=no --color=yes -e '
    using Setfield, InteractiveUtils;
    try
        Base.arg_decl_parts(@which show(stdout, typeof(@lens identity(_))));
    catch err;
        @error "Got an error:" exception = (err, catch_backtrace());
    end;
'
```

prints:

```
Exception handling log message: UndefVarError(:f)
  module=Main  file=none  line=6
  Second exception: UndefVarError(:f)
```

which is handled in

https://github.com/JuliaLang/julia/blob/a0b67dc0c08afe7ccb268057cc87b0d5405f4a63/base/logging.jl#L342-L350

I think all these examples indicate that error reporting system you are using has to be _very_ robust for confidently overloading `show(::IO, ::Type{<:MyType})`.  This is likely not the case in many situations.  So, I think it's reasonable to merge this PR.

davidanthoff:
Thanks a lot for the speedy fix and all the digging to help me understand better what is going on :)

PR #118, opened 2020-01-06
deprecate MultiPropertyLens and @settable

jw3126:
I think `@settable` is an evil no longer necessary and `MulitPropertyLens` was never really the "right design" anyway.

### DIFF
--- a/src/experimental.jl
+++ b/src/experimental.jl
@@ -7,6 +7,12 @@ export MultiPropertyLens
 const NNamedTupleLens{N,s} = NamedTuple{s, T} where {T <: NTuple{N, Lens}}
 struct MultiPropertyLens{L <: NNamedTupleLens} <: Lens
     lenses::L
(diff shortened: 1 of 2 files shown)


src/experimental.jl
```diff
@@ -7,6 +7,12 @@ export MultiPropertyLens
 const NNamedTupleLens{N,s} = NamedTuple{s, T} where {T <: NTuple{N, Lens}}
 struct MultiPropertyLens{L <: NNamedTupleLens} <: Lens
     lenses::L
+    function MultiPropertyLens(lenses::L) where {L <: NNamedTupleLens}
+        @warn """
```
tkf: Maybe use `Base.depwarn`? It's not documented API but I've seen it used outside (e.g., PyCall) so I guess it's OK to use. An upside of using `depwarn` is that you have a better control of it via `--depwarn={yes|no|error}` option.

PR #120, opened 2020-01-12
deprecate ConstIndexLens

### LINKED ISSUE #119: Idea: remove ConstIndexLens and let StaticNumbers.jl handle it
We can use `@lens _[static(1)]` instead of  `@lens _[$1]` <strike>once https://github.com/perrutquist/StaticNumbers.jl/issues/5 is fixed</strike> (edit: fixed by https://github.com/perrutquist/StaticNumbers.jl/pull/6).  I think this is good as it frees `$` from the lens syntax which may be useful for something else (e.g., `@set first($xs) = x` to do `set(xs, (@lens first(_)), x)`).

jw3126:
fix  #119


tkf:
It looks like we need a new version of StaticNumbers.jl.

perrutquist:
StaticNumbers 0.2.2, with inferrable `setindex(::Tuple, _, ::StaticInteger)` has been released now. Thanks for the patch.

tkf:
@perrutquist Thanks for the quick release.

PR #122, opened 2020-01-17
Fix the link to dev version documentation

tkf:
<strike>ATM it's linking to old dev version https://jw3126.github.io/Setfield.jl/latest/intro</strike>


tkf:
Actually, I think Documenter still creates symlink to `/latest`.  But I think it's better to use Documenter's default dev link anyway.

jw3126:
Thanks!

PR #123, opened 2020-01-17
Doctest at-set!

tkf:


jw3126:
LGTM! There is some interesting test fail on nightly:
```julia
IR: Test Failed at /home/travis/build/jw3126/Setfield.jl/test/perf.jl:81

  Expression: uniquecounts(heads_lens) == uniquecounts(heads_hand)

   Evaluated: Dict(:call => 3,:return => 1,:code_coverage_effect => 24,:new => 1) == Dict(:call => 3,:return => 1,:code_coverage_effect => 12,:new => 1)
```
Do you know why and since when there are `:code_coverage_effect` nodes in the IR?

tkf:
Maybe that's due to `--code-coverage=user` option used by `Pkg.test`.  Incidentally, I'm registering a package to avoid just that :smile: : https://github.com/tkf/PerformanceTestTools.jl I think it'll be merged sometime soon: https://github.com/JuliaRegistries/General/pull/7951

jw3126:
Nice, then lets use your package when it is registered!

jw3126:
So I guess `--code-coverage` does introduce a run time counter for each `:code_coverage_effect` statement? So this introduces branches and blocks e.g. simd?

tkf:
> blocks e.g. simd

Yeah, that's pretty much the reason why I created it. See e.g., https://github.com/tkf/BangBang.jl/blob/d0398a14ebc8a9be1b7ee6c02b13de9c36af91cf/test/__test_ir.jl#L17-L30

You can see that I'm counting SIMD instructions like `r"fmul <[0-9]+ x double>"`

PR #129, opened 2020-03-01
add AbstractLensLike

jw3126:
This PR makes it easier to hook into composition and macros for custom lens like objects.

### DIFF
--- a/test/test_new_lens_like.jl
+++ b/test/test_new_lens_like.jl
@@ -1,0 +1,86 @@
+module TestNewLensLike
+
+using Setfield
+using Test
+
+# SEC = SemanticEditorCombinator, see
+# http://conal.net/blog/posts/semantic-editor-combinators
+abstract type AbstractSEC <: Setfield.AbstractLensLike end
+
+struct ComposedSEC{LO, LI} <: AbstractSEC
+    outer::LO
+    inner::LI
+end
+
+Setfield.compose(outer::AbstractSEC, inner::AbstractSEC) = ComposedSEC(outer, inner)
+Setfield.compose(outer::AbstractSEC, inner::Lens) = ComposedSEC(outer, inner)
+Setfield.compose(outer::Lens, inner::AbstractSEC) = ComposedSEC(outer, inner)
+
+struct Constant{V}
+    value::V
+end
+(o::Constant)(x) = o.value
+
+function Setfield.set(obj, sec::AbstractSEC, value)
+    modify(Constant(value), obj, sec)
+end
+
+function Setfield.modify(f, obj, sec::ComposedSEC)
+    let sec=sec, f=f
+        modify(obj, sec.outer) do inner_obj
+            modify(f, inner_obj, sec.inner)
+        end
+    end
+end
+
+struct Elements <: AbstractSEC end
+const elements = Elements()
+
+function Setfield.modify(f, obj, ::Elements)
+    map(f, obj)
+end
+
+@testset "Semantic editor combinators" begin
+    # inspired by https://hackage.haskell.org/package/lens-tutorial-1.0.3/docs/Control-Lens-Tutorial.html
+    molecule = (
+        name="water",
+        atoms=[
+            (name="H", position=(x=1,y=1)),
+            (name="H", position=(x=1,y=2)),
+            (name="O", position=(x=1,y=3)),
+        ]
+    )
+
+    se = @lens _.atoms ∘ $elements ∘ _.position.x
+    res_modify = modify(x->x+1, molecule, se)
+
+    res_macro = @set molecule.atoms ∘ $elements ∘ _.position.x += 1
+    @test res_macro == res_modify
+
+    res_expected = (
+        name="water",
+        atoms=[
+            (name="H", position=(x=2,y=1)),
+            (name="H", position=(x=2,y=2)),
+            (name="O", position=(x=2,y=3)),
+        ]
+    )
+
+    @test res_expected == res_macro
+
+    res_set = set(molecule, se, 4.0)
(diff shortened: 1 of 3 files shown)


test/test_new_lens_like.jl
```diff
@@ -0,0 +1,86 @@
...
+            (name="H", position=(x=1,y=1)),
+            (name="H", position=(x=1,y=2)),
+            (name="O", position=(x=1,y=3)),
+        ]
+    )
+
+    se = @lens _.atoms ∘ $elements ∘ _.position.x
+    res_modify = modify(x->x+1, molecule, se)
+
+    res_macro = @set molecule.atoms ∘ $elements ∘ _.position.x += 1
```
tkf: Very interesting!

I wonder if it's also useful to define SEC-like thing but still with `get` (so it's still a lens).  That is to say, something that returns `[1, 2, 3]` from

```julia
get(molecule, @lens _.atoms ∘ $elements′ ∘ _.position.y)  # y here
```

Then I guess we can get SEC-like behavior with broadcasting?:

```julia
@set molecule.atoms ∘ $elements′ ∘ _.position.y .+= 1
@set molecule.atoms ∘ $elements′ ∘ _.position.x .= 1
```

jw3126: Having a `get` would be great. Any ideas on how it should behave and how to implement it?
* For instance I think it should flatten on composition:
```julia
get([[1]], elements'∘elements')::AbstractVector{Int}
```
* It would be nice if `get` was a cheap operation.
* The user should have a way to overload it

tkf: Do you think it should be auto-flattening to work with broadcasting? Or it should be like that without the consideration of broadcasting?

I actually thought it'd be better to reflect whatever the container referred by the inner lens of the `$elements′` lens:

```julia
get(
    Dict(
        :a => (x = (1im, 2im, 3im), y = nothing),
        :b => (x = (10, 20, 30) .+ 0im, y = nothing)),
    (@lens elements' ∘ _.x ∘ elements' ∘ _.im),
) == Dict(
    :a => (1, 2, 3),
    :b => (10, 20, 30),
)
```

(assuming `map(f, d::Dict) = Dict(k => f(v) for (k, v) in d)`)

> It would be nice if `get` was a cheap operation.

For "large" containers like arrays, I think this would require a lazy wrapper like https://github.com/JuliaArrays/MappedArrays.jl

> The user should have a way to overload it

What aspect you want to overload? Like `map` in the above example?

jw3126: > Do you think it should be auto-flattening to work with broadcasting? Or it should be like that without the consideration of broadcasting?
> 
Yes the broadcasting behavior I had in mind would be more consistent with flattening. Like

```julia
o = Dict(
        :a => (x = (1im, 2im, 3im), y = nothing),
        :b => (x = (10, 20, 30) .+ 0im, y = nothing))
l = (@lens elements' ∘ _.x ∘ elements' ∘ _.im)
@set o \circ $l += 1
```
```julia
o = Dict(
        :a => (x = (2im, 3im, 4im), y = nothing),
        :b => (x = (10, 20, 30) .+ 1im, y = nothing))
```


> I actually thought it'd be better to reflect whatever the container referred by the inner lens of the `$elements′` lens:
> 
> ```julia
> get(
>     Dict(
>         :a => (x = (1im, 2im, 3im), y = nothing),
>         :b => (x = (10, 20, 30) .+ 0im, y = nothing)),
>     (@lens elements' ∘ _.x ∘ elements' ∘ _.im),
> ) == Dict(
>     :a => (1, 2, 3),
>     :b => (10, 20, 30),
> )
> ```
> 
> (assuming `map(f, d::Dict) = Dict(k => f(v) for (k, v) in d)`)
> 
> > It would be nice if `get` was a cheap operation.
> 
> For "large" containers like arrays, I think this would require a lazy wrapper like https://github.com/JuliaArrays/MappedArrays.jl
> 

I guess we can start with the slow variant and switch to lazy later if needed.
> > The user should have a way to overload it
> 
> What aspect you want to overload? Like `map` in the above example?

Yes, if `modify` is the only operation, one probably only needs to overload `map`. If there is also `get`, not sure. Examples I had in mind was traversing a user-defined graph type.

tkf: It looks like the behavior you want is better be implemented via SEC (so no `get`).  I find it actually very interesting that you can get a very powerful tool by dropping `get` from lens but still retaining `modify`.

I think the "SEC-like" lens with non-flattening behavior I proposed might be useful on its own. It is kind of cute that the shape/type of the original container is reflected in the result type. Not that I have a concrete usecase in mind, though.

jw3126: I was also astonished, that dropping  `get` leads to something useful. But now I think there is another perspective from which `get` is the strange guy. I don't see the full picture yet, but I think `SEC` is just a functor (possibly along with some fine print).

* One clue is that you need to overload `map`, which is also what you need to implement for a functor.
* Another is, that all examples in [this blog post](http://conal.net/blog/posts/semantic-editor-combinators) are functors.
* The following looks like the laws that a SEC should satisfy:
```julia
modify(identity, obj, sec) == obj
modify(g  ∘ f, obj, sec) == modify(g, modify(f, obj, sec), sec)
```
These already look a lot like functor laws.

* Given a functor `F` and `f:A->B`, we can define `sec(F)` by:
```julia
modify(f, obj :: F(A), sec(F)) = F(f)(obj)
```
and the functor laws become SEC laws. 

But there are some headaches left with this picture. The main obstacle to setting up a 1:1 correspondence between functors and SEC is to describe for which combinations `modify(f, obj, sec)` is allowed.

PR #132, opened 2020-04-20
Eltype

jw3126:


src/functionlenses.jl
```diff
@@ -14,10 +14,6 @@ set_eltype(::Type{<:Number}, ::Type{T}) where {T} = T
 set_eltype(::Type{<:Array{<:Any, N}}, ::Type{T}) where {N, T} = Array{T, N}
 set_eltype(::Type{<:Dict}, ::Type{Pair{K, V}}) where {K, V} = Dict{K, V}
 set_eltype(obj::Dict, ::Type{T}) where {T} = set_eltype(typeof(obj), T)(obj)
-# This is actually a violation of the lens laws:
-# We have e.g. eltype(@set eltype((1,)), Any) != Any
-set_eltype(obj::Tuple, T::Type) = map(T, obj)
```
tkf: Maybe it's OK if it were `x -> convert(T, x)` instead of `T`? Since `convert` automatically happens with `setfield!` and `setindex!`, it might not be so bad? But yeah, it's kind of ugly...

jw3126: Initially, I thought this is such an innocent violation, but then I produced a bug involving manipulating eltype of `Tuple` and decided to remove this.

tkf: OK. It's better to avoid subtle API, I guess.

PR #136, opened 2020-05-15
clarify == in lens laws #133

### LINKED ISSUE #133: Nitpicking: use isequal instead of == in lens law documentation
`==` in Julia does some interesting things:

```julia
julia> NaN == NaN
false

julia> missing == missing
missing

julia> I(2) == I == I(3)
true
```

jw3126:
Close #133

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -47,16 +47,20 @@ Concrete subtypes of `Lens` have to implement
 These must be pure functions, that satisfy the three lens laws:
 
 ```jldoctest; output = false, setup = :(using Setfield; obj = (a="A", b="B"); lens = @lens _.a; val = 2; val1 = 10; val2 = 20)
-@assert get(set(obj, lens, val), lens) == val
+@assert get(set(obj, lens, val), lens) ≅ val
         # You get what you set.
-@assert set(obj, lens, get(obj, lens)) == obj
+@assert set(obj, lens, get(obj, lens)) ≅ obj
         # Setting what was already there changes nothing.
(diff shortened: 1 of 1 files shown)


src/lens.jl
```diff
@@ -47,16 +47,20 @@ Concrete subtypes of `Lens` have to implement
...
+@assert set(set(obj, lens, val1), lens, val2) ≅ set(obj, lens, val2)
         # The last set wins.
 
 # output
 
 ```
+Here `≅` is an appropriate notion of equality. In most contexts this is simply `==`. But in some
+contexts it might be `===`, `≈`, `isequal` or something else instead. For instance
+`==` does not work in `Float64` context, because `get(set(obj, lens, NaN), lens) == NaN` can never hold.
+Instead `isequal` or `≅(x::Float64, y::Float64) = isequal(x,y) | x ≈ y` are possible alternatives.
```
tkf, line 63: Maybe explicitly say that "equality **or an approximation of it**" so that it's super clear that the "axioms" of equality/equivalence does not have to hold? (or at least one should not expect that arbitrary lens to satisfy that)

```suggestion
Here `≅` is an appropriate notion of equality or an approximation of it. In most contexts
this is simply `==`. But in some contexts it might be `===`, `≈`, `isequal` or something
else instead. For instance `==` does not work in `Float64` context, because
`get(set(obj, lens, NaN), lens) == NaN` can never hold. Instead `isequal` or
`≅(x::Float64, y::Float64) = isequal(x,y) | x ≈ y` are possible alternatives.
```

Issue #62, opened 2019-03-02
Updating more than one field at a time


BenjaminBorn:
Thanks for this very helpful package! I often want to update more than one field in a structure at the same time. Is this somehow possible with this package? I tried

```Julia
using Setfield

struct B
   x::Int
   y::Float64
   z::Bool
end

b = B(1, 2.4, false)

@set b.y, b.z = 2.1, true
```
but that doesn't work. 

If it's not possible yet, would you consider adding such a feature in some form?

Edit: Somewhat unrelated but I just found `@set!`. Should I rather use that than `@set` when replacing values in an existing structure?

jw3126:
Updating multiple values of an object at the same time is currently not possible with `@set`. A lot of people including myself want it. The problem is that somebody actually has to implement it :smile:.

While not as good as versatile as the `@set` macro itself, there is `setproperties`:
```julia
julia> using Setfield: setproperties

julia> setproperties(b, (y=2.1, z=true))
B(1, 2.1, true)
```
Concerning your other question
```julia
@set! b.y = 2.1
```
and 
```julia
b = @set b.y = 2.1
```
are equivalent, so yes I would use `@set!` for replacing existing values.

BenjaminBorn:
Thank you very much for the quick reply. `setproperties` is already very helpful!

singularitti:
@BenjaminBorn FYI, I think what you were asking exists now. Have a look at [Kaleido.jl](https://github.com/tkf/Kaleido.jl), as mentioned in #74.

jw3126:
@singularitti Are you sure? I think there is `@batchlens`, but not `@batchset`?

tkf:
> but not `@batchset`

Right, ATM Kaleido.jl is only for creating lenses and no `@set`-like macros.

singularitti:
> > but not `@batchset`
> 
> Right, ATM Kaleido.jl is only for creating lenses and no `@set`-like macros.

Correct me if I am wrong. But isn't it the `set` statement in your example syntactically equivalent to the `@set b.y, b.z = 2.1, true`?

```julia
julia> using Setfield, Kaleido

julia> lens_batch = @batchlens begin
           _.a.b.c
           _.a.b.d[1]
           _.a.b.d[3] ∘ settingas𝕀
           _.a.e
       end;

julia> obj = (a = (b = (c = 1, d = (2, 3, 0.5)), e = 5),);

julia> get(obj, lens_batch)
[...]

PR #144, opened 2020-06-12
`@set` and `@set!` to change multiple fields

daviehh:
A simple implementation for `@set` and `@set!` to change multiple fields, eg https://github.com/jw3126/Setfield.jl/issues/62.

`@set expr_1 expr_2 ... expr_n`

Some details:

- moved `@assert ex.head isa Symbol` and `@assert length(ex.args) == 2` to a function `set_asserts()` and check via `@assert all(set_asserts.(exprs))`
- get `obj` and `lens` for all expressions via a function `get_obj_lens(exprs)`: checks all expressions are on the same object (such that `@set d.a=1 x.y=2` would fail since it acts on two objects d and x); returns the one object to act on and all lenses
- for the non-overwriting `@set`, from the 2nd expression `expr_2` on, the object would become the destination gensym `dst` otherwise only the last change would apply (only need to be set once). Essentially, set the gensym `dst` with expr_1 and use the new dst as obj for expr_2 through expr_n
- also added test to change multiple fields
- since `@assert` [might be disabled at various optimization levels](https://docs.julialang.org/en/v1/base/base/#Base.@assert), shall we change to use if or is it good as-is?

I originally wanted to keep the original `setmacro` function and call it through all expressions in a new wrapper but since in the non-overwriting mode, dst would be set to a new gensym variable each time `setmacro` gets called, I implemented it inside `setmacro` here.

ex.
```
struct foo
    a
    b
    c
end

d = foo(1,2,3)
```

`q = @set d.a = 10 d.b = 20 d.c = 30`
> foo(10, 20, 30)

`q = @set d.a += 10 d.b *= 20 d.c -= 30`
> foo(11, 40, -27)

`@set! d.a = 11 d.b = -1`
> foo(11, -1, 3)

### DIFF
--- a/docs/src/intro.md
+++ b/docs/src/intro.md
@@ -43,8 +43,15 @@ SpaceShip(Person(:JULIA, 2009), [999999.0, 0.0, 0.0], [0.0, 0.0, 0.0])
 julia> s = @set s.velocity[1] += 1000001
 SpaceShip(Person(:JULIA, 2009), [2.0e6, 0.0, 0.0], [0.0, 0.0, 0.0])
 
-julia> @set s.position[2] = 20
+julia> s = @set s.position[2] = 20
 SpaceShip(Person(:JULIA, 2009), [2.0e6, 0.0, 0.0], [0.0, 20.0, 0.0])
+```
+
+Setting multiple fields also works
+
+```jldoctest spaceship
+julia> @set s.captain.name = :Julia s.captain.age = 1 s.position[1] = 3
+SpaceShip(Person(:Julia, 1), [2.0e6, 0.0, 0.0], [3.0, 20.0, 0.0])
 ```
 
 ## Under the hood
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -180,7 +180,7 @@ function setmacro(lenstransform, exprs::Expr...; overwrite::Bool=false)
         val = (ex.args)[2]
         lens = lenses[idx]
         val = esc(val)
-        if idx >1
+        if idx == 2 && !overwrite
             obj = dst
         end
         ret.args[idx] = if ex.head == :(=)
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -28,8 +28,8 @@ julia> @set t.a.b = 3
 T(T(2, 3), 2)
 ```
 """
-macro set(ex)
-    setmacro(identity, ex, overwrite=false)
+macro set(ex...)
+    setmacro(identity, ex..., overwrite=false)
 end
 
 """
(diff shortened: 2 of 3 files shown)


jw3126:
Thanks a lot for addressing this problem, it is probably the biggest complaint about `Setfield.jl`. It would be awesome to finally have this functionality. So this problem has a couple of dimensions, syntax, semantics, performance, implementation.

## Syntax
I have two small gripes about it:
* The name of the object is redundant, but needs to be repeated:
```julia
@set I_get_tired_writing_this_long_name.more.then.once = 1 I_get_tired_writing_this_long_name.more.then.twice = 2 
```
* If you want to spread the `@set` over multiple lines, you need to put parenthesis
```julia
@set(
    t.a.b=1
    t.b.a = 2
    t.b.c = 3
)
```
Another syntax idea is 
```julia
@set I_get_tired_writing_this_long_name begin
    _.more.then.once = 1 
    _.more.then.twice = 2
end
```
I am open to syntax discussions, though.

## Semantics

The semantics of batch set should be, that it condenses `setproperties` calls in the end. 
So `@set o.a=1 o.b=2` should be equivalent to `setproperties(o, a=1, b=2)` and not 
```julia
o2 = setproperties(o, a=1)
setproperties(o2, b=2)
```
This makes a difference for structs with inner constructors:
```julia
struct ZeroSum
    a
    b
    ZeroSum(a,b) = (@assert a+b = 0; new(a,b))
end

o = ZeroSum(1,-1)
@test (@set o.a = 10 o.b = -10) == ZeroSum(10, -10)
```

## Implementation

I think for each `@set` call there should be a `lens` that corresponds to it. I think for this batched property update [Kaleido.jl](https://tkf.github.io/Kaleido.jl/stable/#Kaleido.jl-1) already has the lens support. We should use a lens like that instead of doing the work in the `setmacro` implementation.
[Here](https://github.com/tkf/Kaleido.jl/issues/27#issuecomment-569458112) is a gist how to implement batch set on top of this lens.

## Performance

The relevant lens infrastructure in `Kaleido.jl` does not infer well. I think this could be fixed at the price of using `@generated` functions in some places. I don't know what @tkf thinks about it. My preferred approach would be to make the relevant `Kaleido.jl` lenses infer and if that works, move it upstream into `Setfield.jl`.
I realize this is more work, then what you probably anticipated when opening this PR. If you are still in, that would be awesome and I am certainly happy to help.

tkf:
> I think this could be fixed at the price of using `@generated` functions in some places.

Yeah, I agree that it is possible and totally OK with going in this direction. But I also think it might be cleaner to do it at macro expansion time where we can do any impure computation. I haven't explored any of these directions yet, though.

daviehh:
Thanks for the really careful thoughts! Yes, I agree that going for a batch-lens approach would be much cleaner. Closing for now, for most applications `setproperties(o, a=1, b=2)` is already really nice. The comment above https://github.com/jw3126/Setfield.jl/pull/144#issuecomment-643582794 is an excellent design rule for possible future implementations. 

Also, thanks for this tool!

Issue #146, opened 2020-07-10
1.0 release


jw3126:
I think the API is pretty stable and the implementation is solid. What do you think about 1.0 @tkf ? Do you have any possible breaking changes in mind, we might want to do?

tkf:
Now that you mention it, I just remembered that that there is an RFC for 2-arg `get(::AbstractDict{K,V}, key) :: Union{Nothing,Some{V}}` API https://github.com/JuliaLang/julia/pull/34821 which would totally ruin Setfield API... :cry:.  Dictionary-like containers would define `get(::ContainerType, key)` while we need `get(obj, ::Lens)`.  This would introduce ambiguities.

(I guess there is some chance that `dict?[key]` become the syntax for `Union{Nothing,Some{T}}` version.  This would be the best-case scenario.  But there is also a big chance that it'd be used for `Union{T,Missing}`.)

I'm not sure what's the best strategy here.  Renaming the function might be the only option but `get` and `set` are such adequate names...  Is there a good name we can use instead of `get`?  Or maybe a better name for https://github.com/JuliaLang/julia/pull/34821?

jw3126:
I actually remember pondering about whether overloading `Base.get` was too daring at the beginning of `Setfield.jl`  :smile:

jw3126:
I was playing with the idea to turn this into a more radical refactoring.

* Replace `Base.get(obj, lens::MyLens)` by `(lens::MyLens)(obj)`
* Use much more duck typing. Get rid of the abstract `Lens` type. A lens can be of any type, it just needs to implement the interface
`Setfield.set(obj, lens, val)` and `lens(obj)`
* In this picture the emphasis is lens as a generalization of `FunctionLens` and not as a generalization of field access

## Pros
* It makes defining function lenses slightly more readable: `Setfield.set(obj, ::typeof(eltype), T) = ...`
* It makes a couple of nice notations more natural, e.g. `@set obj |> first |> eltype = Int`
* I feel that ducktyping + interface is usally a better design then subtying + interface. It's more lightweight to not impose a supertype on downstream users.

## Cons
* Currently we reorder compositions of more than two lenses for performance. We can't do that in the `∘` anymore.
* In haskell lenses are realized as certain functions, so calling a lens in haskell would be a completly different operation then doing it in `Setfield`.
* Pretty printing of composed lenses is a bit more tricky. But I think if `Base` prints `f∘g` as `f∘g` then it should be okay.

Any thoughts @tkf ?

[...]

PR #147, opened 2020-07-15
Rewrite

jw3126:
See #146 List of changes:
* Get rid of `Lens` type, lenses are fully duck typed now
* Replace `get(obj, lens)` by `lens(obj)`
* Replace `set(obj, lens, val)` by `set(lens, obj, val)`
* Replace `modify(f, obj, lens)` by `modify(f, lens, obj)`
* Replace `lens1 ∘  lens2` by `lens2 ∘  lens1`
* Replace `@set!` by `@reset`

@tkf what you think about it. Especially replacing `set(obj, lens, val)` by `set(lens, obj, val)`. In the function lens picture, this feels more natural. Also, it makes a possible future multi object `modify(f, lens, obj, objs...)` easier.

Tests pass locally, except for doctests. Docs are telling lies I plan to fix these when we know what API changes we actually want.
I agree this change is a good opportunity to change the package name.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,13 +10,11 @@ git:
   depth: 99999999
 
 jobs:
-  allow_failures:
-  - julia: nightly
   include:
     - os: osx
-      julia: 1.0
+      julia: nightly
     - stage: "Documentation"
-      julia: 1.4
+      julia: nightly
       os: linux
       script: julia --project=docs --color=yes -e '
           using Pkg;
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ language: julia
 os:
   - linux
 julia:
-  - 1.0
-  - 1.4
-  - 1.5
   - nightly
 notifications:
   email: false
--- a/Project.toml
+++ b/Project.toml
@@ -3,10 +3,12 @@ uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 version = "0.7.0"
 
 [deps]
+CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
 Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+StaticNumbers = "c5e4b96a-f99f-5557-8ed2-dc63ef9b5131"
 
 [compat]
 ConstructionBase = "0.1, 1.0"
@@ -26,4 +28,3 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
 test = ["Test", "Documenter", "PerformanceTestTools", "QuickTypes", "StaticArrays", "BenchmarkTools", "InteractiveUtils", "StaticNumbers"]
-
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Setfield"
 uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-version = "0.7.0"
+version = "0.8.0"
 
 [deps]
 CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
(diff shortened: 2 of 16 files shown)


tkf:
Wow! Thanks a lot for giving a shot at this!

> Especially replacing `set(obj, lens, val)` by `set(lens, obj, val)`. In the function lens picture, this feels more natural.

Isn't this convention usually for `do` block?  Since we can't use anonymous function anyway, I'm not sure if it is a strong motivation on its own.

But maybe it'd let us add `@set(...) do` API like this?

```julia
obj = (a = (b = (c = (1, 2), d = 3,), e = 4,) f = 5)
otherthing = (something = obj, another = nothing)

new_obj = @set(otherthing.something, (40, 50, 20)) do obj
    a = obj.a
    b = a.b
    c = b.c
    return (b.e, obj.f, c[2])
end
```

(Or still with the function `set` but using something like IRTools.jl though that's rather mad. The idea is that `get`/`set` is pretty much like forward/backward passes of the reverse-mode auto-diff.)

> Also, it makes a possible future multi object `modify(f, lens, obj, objs...)` easier.

I think one more nice thing about it is that we can get the curried version `modify(f, lens) = (obj, objs...) -> modify(f, lens, obj, objs...)` more consistently.

But, if we go with this PR, you can also do `set(lens, obj, f(lens.((obj, objs...))...))`, right?  It's a bit tedious to write `lens` twice but I think it's nice overall since you can very clearly see that most of the fields in `objs` are not used this way.  TBH, I'm not _fully_ convinced that `modify(f, lens, obj, objs...)` is the right API.  (That's why I created `DataTools.modifying` outside Setfield.jl.  I wanted to play with it in data processing kind of contexts.)

Hmm...  So `set(lens, obj, val)` is a tough one for me... I'm not super opinionated but I like that in the current `set(obj, lens, val)` the "flow" of the data is consistent (right-to-left, in some sense).

Let me know if you have other consideration on the new `set` API.

> Replace `@set!` by `@reset`

This is nice! (So that we can reserve `@set!` for the in-place version.)

jw3126:
Thanks, you brought up a lot of interesting point. I will focus on `set(lens, obj, val)` vs `set(obj, lens, val)` and leave `modify(f, ...)` for later.
> Hmm... So set(lens, obj, val) is a tough one for me... I'm not super opinionated but I like that in the current set(obj, lens, val) the "flow" of the data is consistent (right-to-left, in some sense).

Interestingly "getting the flow nice" was also for me the main reason to switch to `set(lens, obj, val)` order. My thinking was:
* We humans have inconsistent preferences of preferred flow order. We like:
    - `_.some + _.other      = _.some.other   = get_other(get_some(obj))`
    - `_[some] + _[other]    = _[some][other] = get_other(get_some(obj))`
    - `some(_) + other(_)   = some(other(_)) = get_some(get_other(obj))`
* So `set(obj, lens, val)` is a good fit for property + index flow order, while `set(lens, obj, val)` is better for function order
* I think overall property and index lenses are used more then function lenses. So that is a strong motivation for the `set(obj, lens, val)` order.
* However call syntax forces `(lens, obj)` order on `get`: `lens(obj)`
* It would be a nice consistency if the order was the same for all three operations: `get, set, modify`. I prefer to be able to think in the function order picture all the time over having the property picture most of the time and the function picture sometimes.

Do you see any

tkf:
Thanks for explaining the reasoning behind `set(lens, obj, val)`.  Actually, my thought process was independent of the "internal" order of `lens`.  That is to say, I liked that `set(obj, lens, val)` order has this "alignment"

```julia
obj′ = set(obj,     lens,   val)
@assert    obj′ >>′ lens == val
```

where `>>′` is `get` in current Setfield and `|>` in this PR. The only alternative "flow-compatible" API is

```julia
obj′ = set(val,   lens,    obj)
@assert    val == lens <<′ obj′
```

where `<<′(lens, obj) = obj >>′ lens`.   (But I think this method is very confusing.)

From this perspective, `set(lens, obj, val)` has this "spiral" flow:

```
      ,---.
      |   |
      v   |
set(lens, obj, val)
      |         ^
      |         |
      `---------+
```

That is to say, `lens` has to be in between `obj` and `val`.  I think this is adequate description of lens; it describes a "relationship" of `obj` and `val`.

I guess another way to put it is that (1) `set(obj, lens, val)` is already compatible with function lens view if you consider it in the "`|>`-order" and (2) `val` is at the right location in this view.

jw3126:
Ok, thanks now I get it. I agree data flow compatibility is a nice property. I think with proper `|>` support in the macros that's the best solution.

Issue #154, opened 2021-07-31
`DynamicIndexLens` for `begin`?


phipsgabler:
```julia
julia> @lens _[end]
Setfield.DynamicIndexLens{var"#1#2"}(var"#1#2"())

julia> @lens _[begin]
ERROR: UndefVarError: begin not defined
Stacktrace:
 [1] top-level scope at REPL[6]:1
```

I presume this line:

https://github.com/jw3126/Setfield.jl/blob/544346fce31eb904a951eb651e69fe0fe24490b2/src/sugar.jl#L66

should also contain a test for `:begin`, right?

jw3126:
Oh wow thanks a lot for reporting, interesting that this went unnoticed so long!

PR #155, opened 2021-07-31
Support for `begin` indexing

torfjelde:
This PR adds support for `begin` indexing in the more recent versions of Julia.

Only remaining part is adding tests; could you maybe point me to where I should add some?:)

Related: #154

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -61,25 +61,53 @@ foldtree(op, init, x) = op(init, x)
 foldtree(op, init, ex::Expr) =
     op(foldl((acc, x) -> foldtree(op, acc, x), ex.args; init=init), ex)
 
-need_dynamic_lens(ex) =
-    foldtree(false, ex) do yes, x
-        yes || x === :end || x === :_
-    end
-
-replace_underscore(ex, to) = postwalk(x -> x === :_ ? to : x, ex)
-
-function lower_index(collection::Symbol, index, dim)
-    if isexpr(index, :call)
-        return Expr(:call, lower_index.(collection, index.args, dim)...)
-    elseif index === :end
-        if dim === nothing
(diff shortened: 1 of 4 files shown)


jw3126:
If testing this causes trouble with older julia versions parser, you can create a fresh file. Otherwise you can add tests here:https://github.com/jw3126/Setfield.jl/blob/master/test/test_core.jl#L265

src/sugar.jl
```diff
@@ -61,26 +61,54 @@ foldtree(op, init, x) = op(init, x)
 foldtree(op, init, ex::Expr) =
     op(foldl((acc, x) -> foldtree(op, acc, x), ex.args; init=init), ex)
 
-need_dynamic_lens(ex) =
-    foldtree(false, ex) do yes, x
-        yes || x === :end || x === :_
+@static if VERSION ≥ v"1.5.0-DEV.666"
```
jw3126, line 64: I think instead of duplicating functions, I think I would do something like
```julia
const HAS_BEGIN_INDEXING = VERSION ≥ v"1.5.0-DEV.666"

function need_dynamic_lens(ex)
    ...
    (yes || x === :end || (HAS_BEGIN_INDEXING && x === :begin) || x === :_)
```
and 
```julia
...
function lower_index 
    ...
    elseif (HAS_BEGIN_INDEXING && index === :begin)
            if dim === nothing
                return :($(Base.firstindex)($collection))
            else
                return :($(Base.firstindex)($collection, $dim))
            end
```

torfjelde: That's a good idea; will do that:)

torfjelde:
Added some tests for `begin` :+1:

jw3126 (approved):
LGTM, thanks a lot @torfjelde I will tag a release when #156 is merged.

PR #158, opened 2021-08-01
switch ci to github actions

jw3126:


jw3126:
I will merge so we have CI. Need still to check readme for travis links etc.

PR #162, opened 2021-09-13
Fix #161: Implement == and hash for IndexLens and ComposedLens

### LINKED ISSUE #161: `hash` is "unnecessarily" slow
This is not a very pressing issue, but it caused some confusion over at https://github.com/TuringLang/DynamicPPL.jl/pull/295 and so I figured I'd post the "discovery" here.

So it seems like hashing lenses, in particular compositions of `IndexLens`, is unnecessarily slow compared to what it "represents".

Here's a comparison:

```julia
julia> @benchmark hash($((1, )))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  0.018 ns … 0.043 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.021 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.021 ns ± 0.002 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

     ▅   █  █   ▆  ▂                                   ▂  ▁ ▂
  ▃▁▁█▁▁▁█▁▁█▁▁▁█▁▁█▁▁▁▄▁▁▃▁▁▁▅▁▁▁▄▁▁▄▁▁▁▁▁▁▃▁▁▁▆▁▁▇▁▁▁█▁▁█ █
  0.018 ns    Histogram: log(frequency) by time    0.034 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```
vs.

```julia
julia> @benchmark hash($(@lens(_[1])))
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  8.286 ns … 41.691 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     8.787 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   9.165 ns ±  1.286 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

        █     █                                               
  ▂▁▁▁▁▂█▇▂▂▁▇█▃▂▂▂▃▃▂▂▂▂▃▃▂▂▂▂▂▃▂▂▂▂▂▂▃▂▁▂▂▁▅▆▂▂▂▂▂▂▆▃▂▂▁▂▂ ▃
  8.29 ns        Histogram: frequency by time        10.6 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

This is due to the usage of `objectid` for in the "general" impl of `hash`:

```julia
julia> @benchmark objectid($(@lens(_[1])))
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  6.791 ns … 36.111 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.568 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.854 ns ±  1.268 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

      █  ▆     ▄  ▅                                           
  ▂▁▁▂█▄▂█▇▂▂▄▅█▃▃█▅▃▂▃▂▃▂▃▂█▄▃▂▃▆▃▂▂▂▆▅▂▂▂▄▃▂▂▂▃▂▂▂▂▂▂▂▂▂▂▂ ▃
  6.79 ns        Histogram: frequency by time          10 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

This could be resolved if we implement a custom `hash` for some of the lenses, e.g. `ComposedLens`, `IndexLens`, etc.

phipsgabler:
Now I'm not an expert in these matters, but is seems the bottleneck is not hashing the tuples, but the types or symbols (since [`hash(::Symbol)` falls back to `objectid`](https://github.com/JuliaLang/julia/blob/73a963d37882d70cadb1faf168777514d2b450d1/base/hashing.jl#L27), too).  In the commit without adding type symbols to `hash` (https://github.com/jw3126/Setfield.jl/commit/2ca540462d30de73d3474d80c84fa72aec6115d1):

```julia
julia> @benchmark hash($((1, )))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  0.024 ns … 0.074 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.028 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.029 ns ± 0.003 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                          
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁ ▁
  0.074 ns       Histogram: frequency by time      0.026 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark hash($(@lens(_[1])))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  0.024 ns … 12.903 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.028 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.030 ns ±  0.129 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                           
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁ ▁
  0.063 ns       Histogram: frequency by time       0.027 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

With the symbols (https://github.com/jw3126/Setfield.jl/commit/880daf445f735b2c94fc1b037688f3daf92dfe30):

```julia
julia> @benchmark hash($((1, )))
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
 Range (min … max):  0.024 ns … 0.062 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     0.028 ns             ┊ GC (median):    0.00%
 Time  (mean ± σ):   0.029 ns ± 0.003 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                          
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁ ▁
  0.062 ns       Histogram: frequency by time      0.027 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.

julia> @benchmark hash($(@lens(_[1])))
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):   9.690 ns … 44.821 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     10.423 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   10.712 ns ±  1.721 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

  █                                                            
  █▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▁▁▁  ▁
  15.1 ns         Histogram: frequency by time        10.1 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

Without the symbol, an index lens hashes to the same as the underlying tuple, though:

```julia
julia> hash(@lens(_[1]))
0xf45b969ce741991e

julia> hash((1,))
0xf45b969ce741991e
```

That _could_ be solved by using a magic constant instead, but is that a good idea?

CC @torfjelde

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -9,6 +9,17 @@ export constructorof
 import Base: get, hash, ==
 using Base: getproperty
 
+
+# used for hashing
+function make_salt(s64::UInt64)::UInt
+    if UInt === UInt64
+        return s64
+    else
+        return UInt32(s64 >> 32) ^ UInt32(s64 & 0x00000000ffffffff)
+    end
+end
+
+
 """
     Lens
 
@@ -119,7 +130,9 @@ end
 function ==(l1::ComposedLens{LO, LI}, l2::ComposedLens{LO, LI}) where {LO, LI}
     return l1.outer == l2.outer && l1.inner == l2.inner
 end
-hash(l::ComposedLens, h::UInt) = hash(l.outer, hash(l.inner, h))
+
+const SALT_COMPOSEDLENS = make_salt(0xcf7322dcc2129a31)
+hash(l::ComposedLens, h::UInt) = hash(l.outer, hash(l.inner, SALT_INDEXLENS + h))
 
 """
     compose([lens₁, [lens₂, [lens₃, ...]]])
@@ -179,8 +192,11 @@ end
 struct IndexLens{I <: Tuple} <: Lens
     indices::I
 end
+
 ==(l1::IndexLens{I}, l2::IndexLens{I}) where {I} = l1.indices == l2.indices
-hash(l::IndexLens, h::UInt) = hash(l.indices, h)
+
+const SALT_INDEXLENS = make_salt(0x8b4fd6f97c6aeed6)
+hash(l::IndexLens, h::UInt) = hash(l.indices, SALT_INDEXLENS + h)
 
 Base.@propagate_inbounds function get(obj, l::IndexLens)
     getindex(obj, l.indices...)
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -119,7 +119,7 @@ end
 function ==(l1::ComposedLens{LO, LI}, l2::ComposedLens{LO, LI}) where {LO, LI}
     return l1.outer == l2.outer && l1.inner == l2.inner
 end
-hash(l::ComposedLens, h::UInt) = hash(l.outer, hash(l.inner, hash(:ComposedLens, h)))
+hash(l::ComposedLens, h::UInt) = hash(l.outer, hash(l.inner, h))
 
 """
     compose([lens₁, [lens₂, [lens₃, ...]]])
@@ -180,7 +180,7 @@ struct IndexLens{I <: Tuple} <: Lens
     indices::I
 end
 ==(l1::IndexLens{I}, l2::IndexLens{I}) where {I} = l1.indices == l2.indices
-hash(l::IndexLens, h::UInt) = hash(l.indices, hash(:IndexLens, h))
+hash(l::IndexLens, h::UInt) = hash(l.indices, h)
 
 Base.@propagate_inbounds function get(obj, l::IndexLens)
     getindex(obj, l.indices...)
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -6,7 +6,7 @@ export setproperties
 export constructorof
 
 
-import Base: get
+import Base: get, hash, ==
 using Base: getproperty
 
 """
@@ -99,6 +99,7 @@ struct IdentityLens <: Lens end
 get(obj, ::IdentityLens) = obj
 set(obj, ::IdentityLens, val) = val
 
+
 struct PropertyLens{fieldname} <: Lens end
 
 function get(obj, l::PropertyLens{field}) where {field}
@@ -114,6 +115,11 @@ struct ComposedLens{LO, LI} <: Lens
     outer::LO
     inner::LI
 end
+
+function ==(l1::ComposedLens{LO, LI}, l2::ComposedLens{LO, LI}) where {LO, LI}
+    return l1.outer == l2.outer && l1.inner == l2.inner
+end
+hash(l::ComposedLens, h::UInt) = hash(l.outer, hash(l.inner, hash(:ComposedLens, h)))
 
 """
     compose([lens₁, [lens₂, [lens₃, ...]]])
@@ -173,6 +179,8 @@ end
 struct IndexLens{I <: Tuple} <: Lens
     indices::I
 end
+==(l1::IndexLens{I}, l2::IndexLens{I}) where {I} = l1.indices == l2.indices
+hash(l::IndexLens, h::UInt) = hash(l.indices, hash(:IndexLens, h))
 
 Base.@propagate_inbounds function get(obj, l::IndexLens)
     getindex(obj, l.indices...)
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -206,22 +206,50 @@ end
 end
 
 @testset "equality & hashing" begin
-    # singletons (identity and property lens) are object equal
-    for (l, r) ∈ [
+    # singletons (identity and property lens) are egal
+    for (l1, l2) ∈ [
         @lens(_) => @lens(_),
         @lens(_.a) => @lens(_.a)
     ]
-        @test l === r
+        @test l1 === l2
+        @test l1 == l2
+        @test hash(l1) == hash(l2)
     end
 
     # composite and index lenses are structurally equal
-    for (l, r) ∈ [
+    for (l1, l2) ∈ [
         @lens(_[1]) => @lens(_[1])
         @lens(_.a[2]) => @lens(_.a[2])
+        @lens(_.a.b[3]) => @lens(_.a.b[3])
     ]
-        @test l == r
-        @test hash(l) == hash(r)
+        @test l1 == l2
+        @test hash(l1) == hash(l2)
     end
+
+    # inequality
+    for (l1, l2) ∈ [
+        @lens(_[1]) => @lens(_[2])
+        @lens(_.a[1]) => @lens(_.a[2])
+        @lens(_.a[1]) => @lens(_.b[1])
+    ]
+        @test l1 != l2
+    end
+
+    # Hash property: equality implies equal hashes, or in other terms:
+    # lenses either have equal hashes or are unequal
+    # Because collisions can occur theoretically (though unlikely), this is a property test,
+    # not a unit test.
+    random_lenses = (@lens(_.a[rand(Int)]) for _ in 1:1000)
+    @test all((hash(l2) == hash(l1)) || (l1 != l2)
+              for (l1, l2) in zip(random_lenses, random_lenses))
+
+    # Lenses should hash differently from the underlying tuples, to avoid confusion.
+    # To account for potential collisions, we check that the property holds with high
+    # probability.  
+    @test count(hash(@lens(_[i])) != hash((i,)) for i = 1:1000) > 900
+
+    # Same for tuples of tuples (√(1000) ≈ 32).
+    @test count(hash(@lens(_[i][j])) != hash(((i,), (j,))) for i = 1:32, j = 1:32) > 900
 end
 
 
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -205,6 +205,26 @@ end
     end
 end
 
+@testset "equality & hashing" begin
+    # singletons (identity and property lens) are object equal
+    for (l, r) ∈ [
+        @lens(_) => @lens(_),
+        @lens(_.a) => @lens(_.a)
+    ]
+        @test l === r
+    end
+
+    # composite and index lenses are structurally equal
+    for (l, r) ∈ [
+        @lens(_[1]) => @lens(_[1])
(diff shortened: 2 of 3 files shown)


src/lens.jl
```diff
@@ -173,6 +179,8 @@ end
 struct IndexLens{I <: Tuple} <: Lens
     indices::I
 end
+==(l1::IndexLens{I}, l2::IndexLens{I}) where {I} = l1.indices == l2.indices
+hash(l::IndexLens, h::UInt) = hash(l.indices, h)
```
jw3126, line 183: One minor inconvenience with this is that `hash(@lens _[index]) == hash(index)` holds, e.g. hash forgets about the type. A possible and performant fix would be:
```julia
julia> objectid(IndexLens) # inspiration for the salt
0x8aa710fd8cb95e7e
```
```julia
function hash(l::IndexLens, h::UInt)
    salt = 0x8aa710fd8cb95e7e
    hash(l.indices, hash(salt, h))
end
```
And similar for `ComposedLens`. If there is a performance hit for doing this, we can also leave as is.

test/test_core.jl
```diff
@@ -205,6 +205,26 @@ end
...
+
+    # composite and index lenses are structurally equal
+    for (l, r) ∈ [
+        @lens(_[1]) => @lens(_[1])
+        @lens(_.a[2]) => @lens(_.a[2])
+    ]
+        @test l == r
+        @test hash(l) == hash(r)
+    end
+end
```
jw3126, line 225: Would be nice to also add tests that distinct stuff hashed/compares unequal: `@test @lens(_[1]) != @lens(_[2])`

jw3126 (commented):
Thanks a lot @phipsgabler for looking into this.
> That could be solved by using a magic constant instead, but is that a good idea?

Do you have any specific drawbacks in mind with the magic constant approach? If not I would prefer to have one. (Also for ComposedLens). Possible definition for the constant:
```julia
const SALT_IndexLens = objectid(IndexLens)
```

phipsgabler:
> Do you have any specific drawbacks in mind with the magic constant approach? If not I would prefer to have one. (Also for ComposedLens). Possible definition for the constant:
> 
> ```julia
> const SALT_IndexLens = objectid(IndexLens)
> ```

None other than me having no idea whether this is accepted good practice.  The only objection I could remotely think of is: is it guaranteed that `objectid(IndexLens)` (or even `hash(:IndexLens)`) is constant in a distributed setting?  Is that ever (considered) a problem?

If it's OK with you, I'll go ahead with it.

jw3126:
> None other than me having no idea whether this is accepted good practice. The only objection I could remotely think of is: is it guaranteed that `objectid(IndexLens)` (or even `hash(:IndexLens)`) is constant in a distributed setting? Is that ever (considered) a problem?

I don't know the answer to these either. To be safer, let's just hardcode the following values:
```julia
julia> using Setfield

julia> objectid(Setfield.IndexLens)
0x8aa710fd8cb95e7e

julia> objectid(Setfield.ComposedLens)
0x1c9765b995720e93
```

jw3126:
I also asked about this on [discourse](https://discourse.julialang.org/t/fast-hash-for-custom-type/68135).

JeffreySarnoff:
>  To be safer, let's just hardcode the following values

If by hardcode you mean `const value = objectid(object)`, `objectid(object)` is highly likely to change with different uses of the package and also with different runs of some use of the package.  In a distributed setting, too. It would be trouble to have hash values for the same item in distinct contexts within a single run differ if they were to be checked for equality:

> isequal is the comparison function used by hash tables (Dict). isequal(x,y)
  must imply that hash(x) == hash(y). ...   
  Furthermore, isequal is linked with isless, and they work together to define a
  fixed total ordering, where exactly one of isequal(x, y), isless(x, y), or
  isless(y, x) must be true (and the other two false).

jw3126:
Thanks a lot @JeffreySarnoff for the clarification. By hardcoding I meant 
```julia
const value = 0x8aa710fd8cb95e # objectid(Setfield.IndexLens)
```
> `objectid(object)` is highly likely to change with different uses of the package and also with different runs of some use of the package. 

But the default `hash` implementation uses `objectid` right? So the default `Base.hash` would suffer from the same instabilities?

JeffreySarnoff:
not for all things -- Strings, Tuples of Ints and other bitstypes for example keep their objectids

jw3126:
```julia
struct S{T}
    value::T
end

function hash_cheat(obj, h::UInt)
    hash(obj.value, h)
end

const SALT = 0x8aa710fd8cb95e7e

function hash1(obj, h::UInt)
    hash(obj.value, hash(SALT, h))
end

function hash2(obj, h::UInt)
    hash(obj.value, h+SALT)
end

obj = S((1,))
seed = UInt(1)
using BenchmarkTools

@btime hash_cheat($(Ref(obj))[], $(Ref(seed))[])
@btime hash1($(Ref(obj))[], $(Ref(seed))[])
@btime hash2($(Ref(obj))[], $(Ref(seed))[])

#  1.833 ns (0 allocations: 0 bytes)
#  1.813 ns (0 allocations: 0 bytes)
#  1.833 ns (0 allocations: 0 bytes)
```
I think we should incorporate some type specific salt into our hash and based on the above benchmark, 
it seems there is no performance difference between using `+` or `hash` for the mixing.
So I suggest we use the following implementation:
```julia
function Base.hash(o::IndexLens, h::UInt)
    salt = 0x8aa710fd8cb95e7e #objectid(Setfield.IndexLens)
    hash(o.indices, hash(salt, h))
end
```

JeffreySarnoff:
> it seems there is no performance difference between using + or hash for the mixing.
There is under heavy use; simple benchmarking will not reveal that.

You should define your salt constant outside of and just above the hash function (better practice) and you must provide both a 64bit and a 32bit constant as `hash` uses `UInt` which may be either one.

jw3126:
> > it seems there is no performance difference between using + or hash for the mixing.
> 
> There is under heavy use (Base knows best); simple benchmarking will not reveal that.

Base operates under many constraints like generality and backward compatibility, so "Base knows best" is not obvious to me in this case. But anyway lets do it the Base way:
```julia
function make_salt(s64::UInt64)::UInt
    if UInt === UInt64
        return s64
    else
        return UInt32(s64 >> 32) ^ UInt32(s64 & 0x00000000ffffffff)
    end
end

const SALT_IndexLens = make_salt(0x8aa710fd8cb95e7e) # objectid(Setfield.IndexLens)
Base.hash(o::IndexLens, h::UInt) = hash(o.indices, SALT_IndexLens + h)
```

JeffreySarnoff:
Point taken.  Base knows best about hashing Julia constructs (I should have been more explicit).  A great deal of effort has gone into the hashing approach and its most efficient implementation.

jw3126:
Thanks again @JeffreySarnoff for all your patience and explanations, I appreciate them.

phipsgabler:
Great!  I'll update to the latest version in the next couple of days, plus a few more tests.

test/test_core.jl
phipsgabler: I added some more, check it out.

jw3126 (approved):
LGTM! Can you increase the minor version number @phipsgabler, since this is technically breaking? I will then tag a release immediately.

jw3126:
@phipsgabler thanks again! This is now available in [Setfield 0.8](https://github.com/JuliaRegistries/General/pull/44960#event-5310481912).

tkf:
@jw3126 Sorry if you've already thought through this, but, I think this is actually technically *non*-breaking. The only documented public interface we need to satisfy is `isequal(x,y) implies hash(x)==hash(y)`. So, unless we've already *documented that `hash(lens)` satisfies some additional property*, we can freely re-define `hash` even for patch releases (that'd be annoying, but technically OK) provided it satisfies the above invariance. It's even fine to increment the salt for each release or just use `hash(::Lens) = UInt(0)` (of course, I'm not suggesting this :slightly_smiling_face:).

(I just thought to ramble this, since I noticed JuliaObjects/Accessors.jl#31)

jw3126:
Thanks @tkf So for this PR we already tagged a breaking release. But in the case of Accessors your advise is to only bump the patch version?

tkf:
Yeah, I think Setfield.jl is good as-is. But I don't think Accessors.jl needs a breaking release.

PR #166, opened 2021-11-22
Remove type parameters from == methods of IndexLens and CompositeLens

### LINKED ISSUE #165: Possibility of loosening type-strict equality of indices
The problem I have is this:

```julia
julia> @lens(_[1:10]) == @lens(_[Base.OneTo(10)])
false

julia> @lens(_[1:10]).indices == @lens(_[Base.OneTo(10)]).indices
true
```

arising from the dispatch on a common type in [the definition](https://github.com/jw3126/Setfield.jl/blob/d8938be8f715e570835a8db24951e0a974c90fbe/src/lens.jl#L196):

```julia
==(l1::IndexLens{I}, l2::IndexLens{I}) where {I} = l1.indices == l2.indices
```

I'm asking if there's a possibility of loosening this to

```julia
==(l1::IndexLens, l2::IndexLens) = l1.indices == l2.indices
```

(and the same for `ComposedLens`).  I think this is sensible, since if index values are considered equal, they (hopefully) should also index to equal things.

(I am to [blame for this](https://github.com/jw3126/Setfield.jl/blame/d8938be8f715e570835a8db24951e0a974c90fbe/src/lens.jl#L196) my self anyway :P)

phipsgabler:
Fixes #165.

I'm not quite sure whether this is considered breaking; it changes behaviour of equality for some of the methods...?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Setfield"
 uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-version = "0.8.0"
+version = "0.8.1"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -127,7 +127,7 @@ struct ComposedLens{LO, LI} <: Lens
     inner::LI
 end
 
-function ==(l1::ComposedLens{LO, LI}, l2::ComposedLens{LO, LI}) where {LO, LI}
+function ==(l1::ComposedLens, l2::ComposedLens)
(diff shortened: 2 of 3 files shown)


jw3126:
> I'm not quite sure whether this is considered breaking; it changes behaviour of equality for some of the methods...?

Yeah to me it is a bug fix, but the interpretation that this changes behavior of equality is also valid. I don't feel strongly, but since we need a decision, let's make it a patch release.

jw3126 (approved):
LGTM! Just needs a version bump as discussed above. Thanks @phipsgabler !

jw3126:
Thanks and sorry for the delay!

PR #168, opened 2022-02-12
Allow interpolation of properties

phipsgabler:
This comes up in some VarName use cases in DynamicPPL (cf. https://github.com/TuringLang/AbstractPPL.jl/issues/46).

I simply add a case to check property access with `is_interpolation`.

Any ideas/requirements for more test cases?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Setfield"
 uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-version = "0.8.1"
+version = "0.8.2"
 
 [deps]
 ConstructionBase = "187b0558-2788-49d3-abe0-74a17ed4e7c9"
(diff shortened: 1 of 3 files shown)


jw3126:
Nice! That looks handy. What happens if you try obj.$(f(x) ? An example with a nested attribute access could also be added.

phipsgabler:
You mean like that?

```julia
julia> @lens(_.a[1].$(rand([:x, :y])))
(@lens _.a[1].x)

julia> @lens(_.a[1].$(rand([:x, :y])))
(@lens _.a[1].x)

julia> @lens(_.a[1].$(rand([:x, :y])))
(@lens _.a[1].y)

julia> @macroexpand @lens(_.a[1].$(rand([:x, :y])))
:((identity)((Setfield.compose)((Setfield.PropertyLens){:a}(), (Setfield.IndexLens)((1,)), (Setfield.PropertyLens){rand([:x, :y])}())))
```

jw3126:
I think I will merge this one. If we decide to tweak the parser for https://github.com/TuringLang/AbstractPPL.jl/pull/54 that can be a separate PR. Sounds good @phipsgabler ?

jw3126:
Ah sry @phipsgabler, can you bump the patch version?

Issue #164, opened 2021-10-22
Drop optional dependency


ParadaCarleton:
The single line:
```
setindex(a::StaticArrays.StaticArray, args...) =
            Base.setindex(a, args...)
```
Makes up most of the package's compile time; it could easily be moved to StaticArrays.jl, avoiding the cost of having to use Requires.jl.

jw3126:
mostly solved by #174

PR #174, opened 2022-08-05
Depend on StaticArraysCore.jl; remove Requires.jl dependency

thomvet:
StaticArrays has recently been refactored with the interface contained in StaticArraysCore; a much lighter dependency. This allows to remove the dynamic loading of the StaticArrays package, which has the potential to prevent method invalidations during precompilation of downstream packages.

All tests pass locally (Julia 1.7.2).

Also addresses https://github.com/jw3126/Setfield.jl/issues/164

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 [compat]
 ConstructionBase = "0.1, 1.0"
 MacroTools = "0.4.4, 0.5"
-julia = "1"
+julia = "1.6"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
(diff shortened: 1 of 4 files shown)


Project.toml
```diff
@@ -1,17 +1,16 @@
...
 Future = "9fa8497b-333b-5362-9e8d-4d0656e87820"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 
 [compat]
 ConstructionBase = "0.1, 1.0"
 MacroTools = "0.4.4, 0.5"
-Requires = "0.5, 1.0"
 julia = "1"
```
jw3126, line 14:
```suggestion
julia = "1.6"
```
StaticArraysCore needs this.

Project.toml
```diff
@@ -1,17 +1,16 @@
 name = "Setfield"
 uuid = "efcf1570-3423-57d1-acb7-fd33fddbac46"
-version = "1.0.0"
+version = "1.0.1"
```
jw3126, line 3: I am no semver lawyer, but I think this is more than a patch release, if we bump the julia version, see below.
```suggestion
version = "1.1.0"
```

thomvet:
Thanks for the suggestions; fully agree with both of them.

thomvet:
CI failure was related to Julia 1.0 and 1.3; newest commit switches over to current Julia version (1.7.x), LTS (1.6.x) and nightly.

jw3126 (approved):
LGTM. Thanks a lot @thomvet

PR #177, opened 2022-09-01
Add Invalidations.yml [ci skip] [skip tests]

ranocha:
This is based on https://github.com/julia-actions/julia-invalidations. Adding such checks came up in https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359. I suggest to add this check here since this package is widely used as a dependency.

See also SciML/MuladdMacro.jl#26 and SciML/MuladdMacro.jl#29

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -1,0 +1,40 @@
+name: Invalidations
+
+on:
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


jw3126:
Thanks a lot @ranocha for hunting down invalidation and making sure they don't come back. You are doing a great service to the community. I have a question, I would like to run this action on this branch, before merging. How to manually trigger that?

ranocha:
In SciML, we decided to skip CI for this (since we needed to update ~100 repos). I changed the commit message here to let CI run.

ranocha:
Thus, you should approve workflow runs for this PR and the new GitHub action should run, too.

PR #178, opened 2022-10-27
Adding `hasproperty` overload for lenses

bradcarman:
It's useful to have the `hasproperty(obj, lens)` function in some cases before calling `Setfield.set(obj, lens, val)`

### DIFF
--- a/src/lens.jl
+++ b/src/lens.jl
@@ -259,3 +259,7 @@ struct FunctionLens{f} <: Lens end
 FunctionLens(f) = FunctionLens{f}()
 
 get(obj, ::FunctionLens{f}) where f = f(obj)
+
+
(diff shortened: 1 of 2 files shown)


jw3126:
Thanks for the interesting proposal and PR. I think it would be useful to have a way to find out if a lens is applicable to an object. not just for property lenses. 
In general this is difficult to implement though, especially for user defined lenses. That is why did not do that yet.

phipsgabler:
Indeed this would be nice to have. Such a check is a simplified case of what we call [subsumption](https://github.com/TuringLang/AbstractPPL.jl/blob/3e6f9937f91320c030e0329dcbf44c755c378cfc/src/varname.jl#L268) in Turing/AbstractPPL. It gets complicated quickly... but maybe our implementation can provide some ideas and test cases for Setfield.

